PR #21, opened 2017-08-04
Port most of the buttons from ConsumerWeb

howard-wallis:
Also includes some icons necessary for said buttons.


src/scss/components/_icons.scss
```diff
@@ -0,0 +1,87 @@
+/*
+   Icons
+   ========================================================================== */
+
+
+[class^='c-icon '],
+[class^='c-icon-'],
+[class*=' c-icon '],
```
DamianMullins: Wouldn't 

```css
[class*='c-icon']
```

on its own be sufficient?

[Tested in this CodePen](https://codepen.io/DamianMullins/pen/mMRzbX?editors=1100).

src/img/icons/general/tick-rounded.svg
```diff
@@ -0,0 +1,25 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<svg width="13px" height="14px" viewBox="0 0 13 14" version="1.1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
+    <!-- Generator: sketchtool 41 (35326) - http://www.bohemiancoding.com/sketch -->
+    <title>4B5102F7-6782-4D34-8324-F1639EE7FB4E</title>
```
DamianMullins: With the icons I think we should be stripping out the `title` element in most cases. We do have a Gulp task which optimises the SVG files but it leaves these elements in place as they are sometimes useful/required.

package.json
```diff
@@ -2,7 +2,7 @@
   "name": "@justeat/fozzie",
   "title": "Fozzie – Just Eat UI Web Framework",
   "description": "UI Web Framework for the Just Eat Global Platform",
-  "version": "0.6.0",
+  "version": "0.7.0",
```
ashleynolan: Will need a version bump as v0.7.0 has already been released now

src/scss/components/_icons.scss
ashleynolan: Should be – I think this was only needed before when it was `icon` as it was more likely to be a part of a word.  `c-icon` is much less likely to accidentally be a part of another classname.

src/scss/components/_icons.scss
```diff
@@ -0,0 +1,87 @@
...
+    margin-right: spacing() / 2;
+    width: 13px;
+    height: 23px;
+    background-image: inline('#{$icon-path}/general/tick-rounded.svg');
+    .no-svg & { background-image: url('#{$icon-path}/general/tick.png'); }
+}
+
+.c-icon--tick--blue,
+%c-icon--tick--blue {
+    background-image: inline('#{$icon-path}/general/tick-blue.svg');
```
ashleynolan: This would be better using the SVG spriting as that way we could change the path colour here rather than inline an extra asset.

src/scss/settings/_variables.scss
```diff
@@ -117,9 +117,9 @@ $zIndex: (
 
 // Path vars
 // ==========================================================================
-$img-path                  : '/assets/dist/img'; // Usage: background: url('#{$img-path}/img.jpg') 0 0 no-repeat;
+$img-path                  : './dist/img'; // Usage: background: url('#{$img-path}/img.jpg') 0 0 no-repeat;
```
ashleynolan: These paths should definitely have `!default` after them so they can be overridden potentially within a project.

Ideally, the assets should be copied down from Fozzie to the base project asset folder, otherwise, the image paths will end up being inconsistent between the project and each dependency that references its own assets.

Think we chatted about [this sort of task](https://www.npmjs.com/package/npm-assets) being built into the gulp tasks to help with this.

howard-wallis: Created this issue https://github.com/justeat/gulp-build-fozzie/issues/61 to cover the asset sharing

src/scss/components/_icons.scss
howard-wallis: I chatted with @DamianMullins about how to use SVG spriting here yesterday. Unfortunately this is blocked until [asset sharing](https://github.com/justeat/gulp-build-fozzie/issues/61) is resolved. Shall I put this PR on hold until then or would you rather get just the buttons without icons in?

ashleynolan: There’s no major rush for this PR – we’ll need them for the next pages we work on, but as the asset helper should be incoming this week, I would ask Adam when it’s likely to be merged.

Would keep this PR open until then and just update when we can with the spriting.

src/img/icons/general/tick-rounded.svg
```diff
@@ -0,0 +1,21 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
DamianMullins: Minor: the SVGs are run through an optimisation process in the image gulp tasks, but I think it's tidier if we manually pass assets through [SVGOMG](https://jakearchibald.github.io/svgomg/) once before committing? 

For this SVG the result looks like this:

```html
<svg xmlns="http://www.w3.org/2000/svg" width="13" height="14">
  <g fill="none" fill-rule="evenodd">
    <path fill="#04822C" d="M13 7c0 3.6-3 6.5-6.5 6.5C3 13.5 0 10.5 0 7 0 3.4 3 .5 6.5.5 10 .5 13 3.5 13 7"/>
    <path stroke="#FFF" stroke-width="1.6" d="M9.8 5l-4 4.5-2.7-2" stroke-linecap="round" stroke-linejoin="round"/>
  </g>
</svg>
```

Just means that all those unsightly and unnecessary `id` attributes are stripped out.

package.json
```diff
@@ -50,6 +50,15 @@
     "needs": "^1.1.2",
     "exports": false
   },
+  "assets": [
```
ashleynolan: Can we put this at the very bottom of the `package.json`?  Potentially it could become the biggest part of the file over time and would be easier to see all the other config above the assets being copied.

src/scss/settings/_variables.scss
```diff
@@ -118,9 +118,9 @@ $zIndex: (
 
 // Path vars
 // ==========================================================================
-$img-path                  : '/assets/dist/img'; // Usage: background: url('#{$img-path}/img.jpg') 0 0 no-repeat;
-$icon-path                 : '#{$img-path}/icons';
-$font-path                 : '/assets/dist/fonts';
+$img-path                  : 'fozzie/dist/img' !default; // Usage: background: url('#{$img-path}/img.jpg') 0 0 no-repeat;
```
ashleynolan: How come this folder has changed from `assets/` to `fozzie/`?

howard-wallis: It's really from `dist/` to `fozzie/dist`, as there was no `assets/` directory in Fozzie. But this is due to the asset sharing change as I mentioned in the top comment here: https://github.com/justeat/gulp-build-fozzie/pull/67

I'm looking for a better way to handle this though; currently when I run `gulp` in Fozzie postcss throws out an error: 
```
[16:08:23] -----------------------------------------------------------------------
[16:08:23] Error in plugin: 'gulp-postcss'
[16:08:23] Asset not found or unreadable: fozzie/dist/img/icons/general/tick.svg
[16:08:23] -----------------------------------------------------------------------
```

DamianMullins: Missing the leading slash?

howard-wallis: The problem is that the package name is needed as a prefix to the directories here so that consumers of the package (global-component-library or app_globalweb) can find the `.svg`s in `/assets/dist/imported-assets/fozzie/...` when they run `gulp:css`. But since the `fozzie/` directory doesn't exist within this package, `gulp:css` errors here.

The only way I can find around this so far is to scrap the sorting of imported assets by directory [(removing pkg.name here)](https://github.com/justeat/gulp-build-fozzie/blob/master/tasks/copy.js#L105) but that would allow for potential filename conflicts if two packages export assets with the same path...

howard-wallis: Although, if we are happy to have just the `sprite.svg` rather than individual assets and classes for every icon that would simplify the matter a lot.

ashleynolan: Might be worth catching up together on this one to go through how this works as I’m getting a little confused with the directory structure of the imported assets.

Probably easier to see it in action 👀

howard-wallis: https://github.com/justeat/gulp-build-fozzie/pull/77

howard-wallis:
@ashleynolan @DamianMullins I've updated to the latest gulp tasks and tested that this assets manifest results in the Fozzie icons being pulled into both globalweb and global-component-library. 🎆 

Think this is good to go in?

src/scss/_dependencies.scss
```diff
@@ -65,7 +65,9 @@
  *
  * This is where the majority of our day-to-day styling will take place.  Component styles should be prefixed with `.c-`
  */
-@import 'components/cards.scss';
+@import 'components/buttons';
```
ashleynolan: Didn’t pick up on this before, but buttons should probably be an object (so `o-`) if we are using the definition of objects can be assigned to a tag.

It’s a bit subjective, so don’t particularly mind either way. @DamianMullins any thoughts?

src/scss/components/_icons.scss
ashleynolan: Will this now use the image spriting task in the gulp tasks?  Or are we planning on keeping this inline

src/scss/_dependencies.scss
howard-wallis: 😄  I actually had made them an object first before changing to a component. I wasn't sure since some of them are a little more complex e.g. `btnToggle`

DamianMullins: Just had a look in the UK codebase and they are defined as objects there, I think we should follow their lead and go with `object`?

ashleynolan: Yep – agree

package.json
```diff
@@ -2,7 +2,7 @@
   "name": "@justeat/fozzie",
   "title": "Fozzie – Just Eat UI Web Framework",
   "description": "UI Web Framework for the Just Eat Global Platform",
-  "version": "0.9.0",
+  "version": "0.10.0",
```
ashleynolan: Will need to bump this and the changelog version too ⬆️

PR #24, opened 2017-08-24
v0.9.0 – Adding table styles, utility additions and some small typographic changes

ashleynolan:
### Added
- Table Styling
- `$border-radius` variable added to standardise the radius of rounded components.
- Utility class `u-lightenBg` added so that a lighter background colour can be applied to a page or section of a page.

### Changed
- Better list item spacing.  By default now adds spacing between `li`s (which can be removed by applying the `u-unstyled` class).
- More sensible typographic spacing based on baseline spacing.


src/scss/objects/_tables.scss
```diff
@@ -1,4 +1,105 @@
...
  * Objects > Tables
  * =================================
+ * File Index:
+ * - Table specific variables
+ * - Default table styles
+ * – Table modifiers:
+ *   - .table--condensed
+ *   - .table--bordered
+ *   - .table--rounded
+ *   - .table--striped
```
DamianMullins: I can only see `.table--striped` in the styles below?

ashleynolan: Yep – have adjusted that now.  It’s all default styling for tables now.

PR #25, opened 2017-08-25
v0.10.0 – Content classes updated, jumbo text line-height, and custom bullets

ashleynolan:
### Added
- Custom bullets on unordered lists.
- `u-bordered` class for adding a border around a block.
- Added narrow view type sizes for headings

### Changed
- Updated line-height for jumbo homepage text.
- Updated the `l-content` class so that it can contain a `header` and `main` that are appropriately spaced for content pages.
- Updated packages

### DIFF
--- a/src/scss/base/_typography.scss
+++ b/src/scss/base/_typography.scss
@@ -67,9 +67,13 @@ h6 {
 // For example: h3.h1 or h3.alpha
 h1,
 .alpha {
-    @include font-size(jumbo);
+    @include font-size(jumbo, true, narrow);
     margin-top: $baseline * 2;
     margin-bottom: $baseline;
+
+    @include media('>mid') {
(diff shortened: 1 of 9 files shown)


src/scss/base/_typography.scss
```diff
@@ -67,18 +67,26 @@ h6 {
 // For example: h3.h1 or h3.alpha
 h1,
 .alpha {
-    @include font-size(jumbo);
+    @include font-size(jumbo, true, narrow);
```
DamianMullins: Question; what are the last two parameters?

ashleynolan: (font-key, line-height output (boolean), subkey)

So here it’s saying, get me the font-size of jumbo for the narrow subkey, and output the line-height as well.

PR #29, opened 2017-09-26
v0.14.0 — Removing icons, and tidying up project files and dependencies.

DamianMullins:
### Changed
- Tidied up gulp setup and config.
- Moved build packages in `devDependencies`.

### Removed
- Removed icons — they now live in the [`f-icons`](https://github.com/justeat/f-icons) module.
- Removed `require-dir` package.
- Removed placeholder JavaScript.


src/js/fozzie.js
```diff
@@ -1,7 +0,0 @@
-import module from './module';
-
-const fozzie = () => 'Wocka wocka wocka!';
-
-module();
```
ashleynolan: Is it worth keeping this in for is we want to move some of the more common JS into the base of fozzie?

DamianMullins: I think it's OK for now. 

I've created a related issue (#30) where I've linked to the commit which removed the JS for reference when we come to add some JS back in.

PR #31, opened 2017-10-03
v0.15.0 — Updated icon styles.

DamianMullins:


src/scss/_dependencies.scss
```diff
@@ -67,7 +67,6 @@
  * This is where the majority of our day-to-day styling will take place.  Component styles should be prefixed with `.c-`
  */
 @import 'components/cards';
-@import 'components/icons';
```
ashleynolan: Should this be replaced by the reference to the eyeglass Sass in `f-icons`?

DamianMullins: As the `f-icons` module contains images to pull into the consuming project, they would need to be imported at that level instead as `f-copy-assets` only copies one level deep into the `node_modules` folder.

PR #35, opened 2017-10-26
CPD-8528 Truncate mixin is added

bulutcy:


ashleynolan:
Can you add the changes to the [CHANGELOG](https://github.com/justeat/fozzie/blob/master/CHANGELOG.md)?

Just copy the format in that file.

bulutcy:
Updated, please review.

src/scss/objects/_buttons.scss
```diff
@@ -12,7 +12,7 @@ $btn-default-bgColor             : $grey--lightest !default;
 $btn-default-weight              : 700 !default;
 $btn-default-height              : 2.6 !default;
 $btn-default-font-size           : mid !default;
-$btn-default-font-family         : inherit !default;
+$btn-default-font-family         : 'Ubuntu';
```
DamianMullins: Should keep `!default` so that the value can be overridden if required.

package.json
```diff
@@ -30,7 +30,7 @@
     "node": ">=4.0.0"
   },
   "dependencies": {
-    "@justeat/fozzie-colour-palette": "^0.8.0",
+    "@justeat/fozzie-colour-palette": "^0.9.0",
```
DamianMullins: This version of `fozzie-colour-palette` needs to be published now that it has been merged. The Travis build is failing because the version in the Yarn lockfile does not match.

bulutcy:
Color palette is published, this is ready to be merged.

ashleynolan:
:shipit: All good to publish

PR #37, opened 2017-11-01
Added u-uppercase utility class

oleksiishevchukje:
Should we bump MINOR version or PATCH is enough for such change?


ashleynolan:
@oleksiishevchukje Patch version is fine for that change 👍

PR #39, opened 2017-11-06
Focus outline emulation using box-shadow

sedan:
Looks better than default outline on elements with rounded corners


sedan:
@ashleynolan
Bumped the version number and updated CHANGELOG, please take a look

PR #41, opened 2017-11-07
Allow Fozzie to output font sizes in EMs if specified.

lsphillips:
I feel Fozzie is being too opinionated by using just REM. This unit is fine when you control everything on the page, however, when developing a web component (like Partner Centre Navigation) using Fozzie that will be dropped on a page that may not be using Fozzie (like Partner Insights) or has a different root font size.

This PR allows the `font-size` mixin to output EM when specified. The default is to still output REM, so this change is backward compatible.

### DIFF
--- a/src/scss/tools/mixins/_type.scss
+++ b/src/scss/tools/mixins/_type.scss
@@ -17,10 +17,10 @@
     // if the $key passed in is actually just a font-size, then output it (may want to instead force a sass error saying to use the type map?)
     @if type-of($key) == 'number' {
         @if $rem == true {
-            @include rem(font-size, $key, $rem);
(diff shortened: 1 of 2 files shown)


src/scss/tools/mixins/_type.scss
```diff
@@ -1,19 +1,27 @@
 // ==========================================================================
 // Typography mixins
 // ==========================================================================
-// REM
```
ashleynolan: Any reason for removing this comment?

ashleynolan:
Happy to merge this in – all that you need to add is an entry to the CHANGELOG (in the root of the project) and bump the version number in the package.json so we can republish.

PR #47, opened 2018-01-26
CPD-11417 – Switch to using `babel-preset-env`

ashleynolan:
### Changed
- Updated `browserlist` in `package.json`
- Updated to use `babel-preset-env`

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,7 +2,7 @@
   "name": "@justeat/fozzie",
   "title": "Fozzie – Just Eat UI Web Framework",
   "description": "UI Web Framework for the Just Eat Global Platform",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -70,8 +70,9 @@
   "browserslist": [
     "> 5%",
     "last 2 versions",
-    "ie > 8",
-    "Safari >= 8"
+    "ie > 10",
```
DamianMullins: Sorry to be annoying but I've just noticed that you can remove this line and the same browsers will be targeted.

ashleynolan: 😡Done 😉

PR #54, opened 2018-02-06
Gs 319 add breadcrumbs component

80sgreek:


src/scss/components/_breadcrumbs.scss
```diff
@@ -0,0 +1,54 @@
...
+    width: 100%;
+    @include font-size(mid, false);
+    line-height: 3;
+}
+
+.c-breadcrumb-item {
+    float: left;
+    color: $grey--dark;
+
+    &--home {
```
ashleynolan: Prefer to include the full selector rather than using `&` shorthands, just as it's easier to search or find and replace should someone need to find selectors later.

PR #58, opened 2018-02-14
Update to replace css inlined svgs that are breaking build

80sgreek:
Update to replace css 'inlined' svgs with svg icons for the breadcrumb component.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
-v0.27.0
+v0.28.0
 ------------------------------
 *February 14, 2018*
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,14 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 v0.27.0
 ------------------------------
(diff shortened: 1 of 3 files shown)


kevinrodrigues:
@80sgreek Sorry to be a pain but could we relate this back to a jira ticket please?
https://jira.just-eat.net/browse/CPD-11917

CHANGELOG.md
```diff
@@ -3,6 +3,14 @@
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+v0.27.0
```
kevinrodrigues: `0.28.0`

src/scss/components/_breadcrumbs.scss
```diff
@@ -32,23 +27,12 @@ $breadcrumb-chevron-point-right : -90deg;
     color: $grey--darkest;
     font-weight: $breadcrumb-font-weight;
 
-    //positioning for rotated chevron in :after
```
ashleynolan: Just to get up to speed on this – doesn't fozzie have a dependency on `f-icons`and so the SVG should be there for this inlining?

80sgreek: Not at the moment, this is the quickest way to get it building again

ashleynolan: Ah, ok – guessing there's a task to add this back in then off the back of this?

PR #60, opened 2018-02-16
CPD-11915: Renamed fozzie index scss file to fozzie.

kevinrodrigues:
This is required so the CDN task outputs a named file some more details here:
https://jira.just-eat.net/browse/CPD-11915


ashleynolan:
One thing I just thought of that's worth checking – eyeglass hooks into the `index.scss` file.  Might be worth checking that when it's imported we don't have to [change this](https://github.com/justeat/fozzie/blob/master/package.json#L52) now the filename is changing

kevinrodrigues:
@ashleynolan good catch! Will check it out.

kevinrodrigues:
@ashleynolan I read into Eyeglass yesterday and looked that their code. I tested this locally too and it seems that it doesn't matter what the base scss file is called as long as it's a valid one scss file. I did test this locally [here](https://github.com/justeat/global-component-library) I changed the index.scss file to fozzie.scss and the UI documentation compiled fine. Let me know if you need me to test it anywhere else? :)

ashleynolan:
@kevinrodrigues Sounds good.  Main thing to check once this is released is that globalweb pulls in the fozzie styles as normal (as that site, OrderWeb and Payments do the same import).  If it works on the docs though, it should work fine on those, just worth a double check when this gets published.

kevinrodrigues:
@ashleynolan Just to update this PR. I spent some time this morning just making sure changing index.scss > fozzie.scss works ok in globalweb. I installed v0.29.0 and then checked the FE and also the node_modules folder to make sure it was renamed correctly and that it still pulls through the bsae Fozzie styles (Checked it against the ). Looks all good. 💃

ashleynolan:
@kevinrodrigues Awesome 💯 🎯

PR #61, opened 2018-02-20
(CPD-8021) CSS breakpoints

xander-marjoram:
**Browsers tested:**
- [x] Chrome
- [x] Firefox
- [x] Safari


gulpfile.js
```diff
@@ -2,12 +2,4 @@ const gulp = require('gulp');
...
-build(gulp, {
-    js: {
-        files: {
-            main: {
-                srcPath: 'fozzie.js'
-            }
-        }
-    }
-});
+build(gulp, {});
```
ashleynolan: This can just be `build(gulp);`

src/js/index.js
```diff
@@ -0,0 +1,7 @@
+// All helper functions will be imported here, so that they can all be exported within one object.
+import { getBreakpoints, currentScreenWidth } from './modules/breakpointHelper';
+
+export {
+    getBreakpoints,
+    currentScreenWidth
```
ashleynolan: Would make this function name an action, so something like `getCurrentScreenWidth`

src/js/modules/breakpointHelper.js
```diff
@@ -0,0 +1,56 @@
+/**
+ * @overview Breakpoint handler
+ *
+ * @module breakpointHelper
+ */
+
+export const getBreakpoints = () => {
```
ashleynolan: Would be good to comment these functions, so that they have a brief description of what they do and what they return

src/scss/settings/_breakpoints.scss
```diff
@@ -0,0 +1,16 @@
+@function populate-breakpoints() {
```
ashleynolan: I'd probably put this file in the root of `scss/tools`, as it's more of a tool than a setting.

src/scss/tools/functions/_units.scss
```diff
@@ -57,3 +57,18 @@
         @return $number;
     }
 }
+
+@function ko-px($emVal, $base: $font-size-base) {
```
ashleynolan: Would be good to rename these so that they are fozzie related (as eventually we're going to port the functions from the kickoff utils and rename them).

So maybe `em-to-px` and just drop the `ko-` from the function below so `map-to-px`

CHANGELOG.md
```diff
@@ -3,6 +3,22 @@
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+
+v0.29.0
```
ashleynolan: Need to bump this to `v0.30.0` as another PR just got merged which was 29.

src/js/modules/breakpointHelper.js
```diff
@@ -0,0 +1,56 @@
...
+    return parseInt(pixels.slice(0, -2), 10);
+}
+
+export const currentScreenWidth = () => {
+    const currentWidth = window.innerWidth;
+
+    const breakpoints = getBreakpoints();
+    const bps = Object.entries(breakpoints).reverse();
+
+    for (let i = 0; i < bps.length; i++) {
```
DamianMullins: Use `forEach()`?

src/js/test/index.test.js
```diff
@@ -0,0 +1,105 @@
...
+        expect(breakpoints.wide).not.toBeUndefined();
+        expect(breakpoints.huge).not.toBeUndefined();
+    });
+
+    it('should return breakpoints in pixels', () => {
+        // Act
+        const breakpoints = getBreakpoints();
+
+        // Assert
+        expect(breakpoints.narrow.slice(-2)).toBe('px');
```
DamianMullins: Just test for the full value i.e. `414px` rather than calling slice throughout.

src/js/test/index.test.js
```diff
@@ -0,0 +1,105 @@
...
+});
+
+describe('currentScreenWidth', () => {
+
+    it('returns a string if breakpoints are defined correctly', () => {
+        // Arrange
+        TestUtils.setBodyHtml(`
+        <style>
+            .c-screen-sizer {
+                display: none;
```
DamianMullins: Do you need the `display: none;` part in the tests?

src/js/test/index.test.js
```diff
@@ -0,0 +1,105 @@
...
+        // Assert
+        expect(typeof breakpoints).toBe('object');
+    });
+
+    it('should return multiple breakpoints', () => {
+        // Act
+        const breakpoints = getBreakpoints();
+
+        // Assert
+        expect(Object.entries(breakpoints).length > 1);
```
DamianMullins: Test that the length equals `4`?

src/js/modules/breakpointHelper.js
```diff
@@ -0,0 +1,56 @@
...
+        .replace(/"/g, '')
+        .replace(/'/g, '')
+        .split(',');
+
+    if (breakpoints.length === 1 && breakpoints[0] === '') {
+        // When there is no content
+        return output;
+    }
+
+    breakpoints.forEach(breakpoint => {
```
DamianMullins: Can use `reduce()` and return the value here:

```js
return breakpoints.reduce((prev, current) => {
    const [breakpointName, breakpointValue] = current.split(':');
    prev[breakpointName] = breakpointValue;
    return prev;
}, {});
```

src/js/modules/breakpointHelper.js
```diff
@@ -0,0 +1,56 @@
...
+        const [breakpointName, breakpointValue] = breakpoint.split(':');
+        output[breakpointName] = breakpointValue;
+    });
+
+    return output;
+};
+
+function pixelStringToInt (pixels) {
```
DamianMullins: Use arrow function to keep the style consistent with other functions:

```js
const pixelStringToInt = pixels => parseInt(pixels.slice(0, -2), 10);
```

DamianMullins: Looking at this code, you don't actually need the `slice()` part, `parseInt()` will ignore non-numerical characters, which means you can just use parseInt in the code below.

src/js/modules/breakpointHelper.js
```diff
@@ -0,0 +1,56 @@
...
+
+export const currentScreenWidth = () => {
+    const currentWidth = window.innerWidth;
+
+    const breakpoints = getBreakpoints();
+    const bps = Object.entries(breakpoints).reverse();
+
+    for (let i = 0; i < bps.length; i++) {
+        // Loops through the breakpoints (in descending order) and returns the first one that is smaller than the current window width.
+        const pixels = pixelStringToInt(bps[i][1]);
```
DamianMullins: Use `parseInt()` directly here.

src/js/modules/breakpointHelper.js
xander-marjoram: I'd argue that's actually harder to interpret than a simple for-each loop.

src/js/test/index.test.js
xander-marjoram: Nope. Will remove 👋

src/js/test/index.test.js
xander-marjoram: This also means I can remove the next test 🕺

src/scss/settings/_breakpoints.scss
xander-marjoram: Will do 🚯

CHANGELOG.md
xander-marjoram: 🥈

src/js/test/index.test.js
```diff
@@ -0,0 +1,92 @@
...
+        // Assert
+        expect(typeof breakpoints).toBe('object');
+    });
+
+    it('should return four breakpoints', () => {
+        // Act
+        const breakpoints = getBreakpoints();
+
+        // Assert
+        expect(Object.entries(breakpoints).length === 4);
```
kevinrodrigues: `expect(Object.entries(breakpoints)).toBe(4)`

src/js/modules/breakpointHelper.js
```diff
@@ -0,0 +1,62 @@
...
+    // Append hidden element to body
+    const screenSizer = document.createElement('div');
+    screenSizer.classList.add('c-screen-sizer');
+
+    document.body.appendChild(screenSizer);
+
+    // It should have a 'content' property containing the breakpoints
+    const breakpoints = window.getComputedStyle(document.querySelector('.c-screen-sizer'))
+        .getPropertyValue('content')
+        .replace(/"/g, '')
```
DamianMullins: Do you need these two lines? Would there ever be any quotes in the value? If you do need them you should be able to reduce to a single line: `.replace(/['"]+/g, '')`

src/js/test/index.test.js
DamianMullins: Use `toBe(4)` instead of `===`.

DamianMullins: Jinx! 👾

src/js/test/index.test.js
```diff
@@ -0,0 +1,92 @@
...
+    });
+
+    it('should return breakpoints with expected names', () => {
+        // Act
+        const breakpoints = getBreakpoints();
+
+        // Assert
+        expect(breakpoints.narrow).not.toBeUndefined();
```
kevinrodrigues: You could use `toBeDefined` so you don't have to call `not`

gulpfile.js
```diff
@@ -2,12 +2,4 @@ const gulp = require('gulp');
 const { build } = require('@justeat/gulp-build-fozzie');
 
 // Grab the Fozzie build tasks
-build(gulp, {
-    js: {
```
kevinrodrigues: Would it make sense to have this instead of removing it? Reason being is that for my CDN task I need the JS file to be a named file i.e `fozzie.js` or `fozzie.min.js` as apposed to just `scripts.js` etc. Teamcity has a build step that outputs the css and js files to S3, currently I have another repo outputting a scripted JS file (with the same name) so to avoid confusion I was going to remove it (after the discussion in hipchat earlier today) but as your add JS now it would be nice to add the below so it does it automatically ☺️ 

```javascript
build(gulp, {
    js: {
        files: {
            main: {
                srcPath: 'index.js',
                distFile: 'fozzie.js'
            }
        }
    }
});
```

src/js/modules/breakpointHelper.js
ashleynolan: They are needed, as the value returned is a bit of a weird format – might be able to reduce it down though

src/js/test/index.test.js
xander-marjoram: @kevinrodrigues `expect(Object.entries(breakpoints).length).toBe(4)` ?

kevinrodrigues: Sorry missed the `.length` :)

src/js/modules/breakpointHelper.js
DamianMullins: It's OK for now but I think we should be pushing to use functions like these a bit more in the codebase, adding a short comment could explain what the `reduce` does for anyone unfamiliar?

package.json
```diff
@@ -2,7 +2,8 @@
   "name": "@justeat/fozzie",
   "title": "Fozzie – Just Eat UI Web Framework",
   "description": "UI Web Framework for the Just Eat Global Platform",
-  "version": "0.29.0",
+  "version": "0.30.0",
+  "main": "dist/index.js",
```
DamianMullins: Would this now be `fozzie.js`?

kevinrodrigues: I thought that too, but because that's the main entry point (`index.js`) it shouldn't need updating. :)

gulpfile.js
```diff
@@ -6,7 +6,8 @@ build(gulp, {
     js: {
         files: {
             main: {
-                srcPath: 'fozzie.js'
+                srcPath: 'index.js',
```
kevinrodrigues: My bad here but on second thoughts you could actually remove this line  as index.js is the default :)

package.json
DamianMullins: @xander-marjoram you're going to need to update the [`prepare` npm script](https://github.com/justeat/fozzie/blob/418ace8f26e8fba20cc7d823ce2f533eab53618e/package.json#L60) so that the JS is compiled and output into the `dist` folder. 

We use the [`concurrently` npm module](https://github.com/justeat/f-header/blob/20ac75eb309f373c2a8886f974a98d40168abbe9/package.json#L56) to run multiple tasks in parallel.

DamianMullins: @kevinrodrigues the file which is output into the `dist` directory is called `fozzie.js` now though so the JS wouldn't be accessible?

DamianMullins: Have just seen your other comment below @kevinrodrigues, so this line should be fine once the `prepare` script has been updated 🍃

src/js/modules/breakpointHelper.js
```diff
@@ -0,0 +1,64 @@
...
+
+    // Append hidden element to body
+    const screenSizer = document.createElement('div');
+    screenSizer.classList.add('c-screen-sizer');
+
+    document.body.appendChild(screenSizer);
+
+    // It should have a 'content' property containing the breakpoints
+    const breakpoints = window.getComputedStyle(document.querySelector('.c-screen-sizer'))
+        .getPropertyValue('content')
```
DamianMullins: Can you update the PR description to include a list of the browsers tested, please? Asking as I don't think I've used this function before so want to be sure we aren't breaking anything 🦌

src/js/modules/breakpointHelper.js
```diff
@@ -0,0 +1,64 @@
...
+        bps.unshift([key, breakpoints[key]]);
+    });
+
+    for (let i = 0; i < bps.length; i++) {
+        // Loops through the breakpoints (in descending order)
+        // returning the first one that is narrower than currentWidth.
+
+        const breakpointWidth = parseInt(bps[i][1], 10); // This also strips the 'px' from the string
+
+        if (i === bps.length - 1) {
```
kevinrodrigues: `if (i === bps.length - 1 || currentWidth > breakpointWidth)` Should do the same? that way it's all on one line :)

package.json
```diff
@@ -41,7 +44,7 @@
     "babel-preset-env": "^1.6.1",
     "danger": "^3.0.5",
     "gulp": "^3.9.1",
-    "npm-scripts-info": "^0.3.6"
```
DamianMullins: `npm-scripts-info` is still in the `yarn.lock` file, have you checked in the latest changes to that file?

gulpfile.js
```diff
@@ -6,7 +6,7 @@ build(gulp, {
     js: {
         files: {
             main: {
-                srcPath: 'fozzie.js'
+                distFile: 'index.js'
```
kevinrodrigues: Could this be fozzie.js please :)

kevinrodrigues:
Some conflicts here @xander-marjoram

src/js/modules/breakpointHelper.js
xander-marjoram: Updated 👍 I was getting an error with `checkForUser is not a function` (from f-header), I just had to update the f-header version in globalweb to latest.

package.json
```diff
@@ -2,9 +2,11 @@
   "name": "@justeat/fozzie",
   "title": "Fozzie – Just Eat UI Web Framework",
   "description": "UI Web Framework for the Just Eat Global Platform",
-  "version": "0.30.0",
+  "version": "0.32.0",
+  "main": "dist/js/index.js",
```
DamianMullins: This needs to be `fozzie.js` now right?

.eslintrc
```diff
@@ -1,17 +1,7 @@
...
-
     "rules": {
-        "jest/no-disabled-tests": "warn",
-        "jest/no-focused-tests": "error",
-        "jest/no-identical-title": "error"
-    },
-
-    "env": {
-        "jest/globals": true
+        "import/prefer-default-export": "off"
```
DamianMullins: Do you need this rule?

package.json
ashleynolan: Nope – it's a bit confusing at the moment as we have two resolved entry points.

`fozzie.js` is the name of the file the gulp build will create.  `index.js` is the file used as the entry point when imported as a module.

DamianMullins: Gotcha.

PR #62, opened 2018-02-22
GS-375 Port media element over to Fozzie

80sgreek:
Direct port of media element with mixin updates where necessary for compatibility.

### DIFF
--- a/src/scss/components/_media-element.scss
+++ b/src/scss/components/_media-element.scss
@@ -7,19 +7,16 @@
  *
  */
 
-$mediaElement-img-width: 50px !default;
-$mediaElement-img-height: 50px !default;
+$mediaElement-img-width: 50px;
 
-$mediaElement-img-width--mid: 78px !default;
-$mediaElement-img-height--mid: 78px !default;
+$mediaElement-img-width--mid: 78px;
 
-$mediaElement-img-width--wide: 67px !default;
-$mediaElement-img-height--wide: 67px !default;
+$mediaElement-img-width--wide: 67px;
 
 
 // mediaElement module
 // Usual of having an image to the left and content to the right
-.mediaElement {
+.c-mediaElement {
     display: flex; //defaults to row
     align-items: flex-start;
     width: 100%;
@@ -32,21 +29,21 @@ $mediaElement-img-height--wide: 67px !default;
 }
 
 // stacks the elements (across all viewports)
-.mediaElement--fullstack {
+.c-mediaElement--fullstack {
     flex-direction: column;
     align-items: center;
     justify-content: center;
 }
 
 // stacks the elements (only on narrow)
-.mediaElement--stack {
+.c-mediaElement--stack {
     @include media('<mid') {
         flex-direction: column;
         align-items: center;
     }
 }
 
-    .mediaElement-img {
+    .c-mediaElement-img {
         line-height: 0;
         margin-right: 5%;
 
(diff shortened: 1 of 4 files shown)


src/scss/components/_media-element.scss
```diff
@@ -0,0 +1,134 @@
...
+ * Media element
+ * ===================================
+ * Example – SERP – Restaurant logos
+ *
+ * A common pattern that sees an image followed by a block of information.
+ *
+ */
+
+$mediaElement-img-width: 50px !default;
+$mediaElement-img-height: 50px !default;
```
kevinrodrigues: Do you need this variable and the other height variants? They don't seem to be used anywhere here.

src/scss/components/_media-element.scss
```diff
@@ -0,0 +1,134 @@
...
+                display: none;
+            }
+        }
+    }
+
+    .mediaElement-text {
+        margin-top: 0;
+        @include font-size(small, false);
+    }
+    .mediaElement-text--large {
```
kevinrodrigues: More of a question, should there be a line break before this selector?

80sgreek:
Just realised I need to update the format to include a 'c-' at the beginning of each class.

src/scss/components/_media-element.scss
80sgreek: @kevinrodrigues good point :)

I'll remove these, maybe they were added as an indicator of 'square logos'

80sgreek:
@kevinrodrigues I've also removed the !default flag from the width variables. They can be added back in the future when/if needed.

src/scss/components/_media-element.scss
```diff
@@ -0,0 +1,132 @@
...
+ * Media element
+ * ===================================
+ * Example – SERP – Restaurant logos
+ *
+ * A common pattern that sees an image followed by a block of information.
+ *
+ */
+
+$mediaElement-img-width: 50px;
+
```
ashleynolan: Super nit picky – any need for the spaces between these vars?

src/scss/components/_media-element.scss
```diff
@@ -0,0 +1,132 @@
...
+$mediaElement-img-width--mid: 78px;
+
+$mediaElement-img-width--wide: 67px;
+
+
+// mediaElement module
+// Usual of having an image to the left and content to the right
```
ashleynolan: This line doesn't read quite right.  Maybe should be `Usually consists of having…`

ashleynolan:
@tommaton @80sgreek The way we decided was least confusing (as objects and components are so similar) was if objects only apply to native HTML tags.  So objects will be pretty minimal – things like `o-btn`, `o-input` that kind of thing.  Everything else should be a component.

ashleynolan:
Looks good – just a couple of tiny documentation comments but otherwise good to merge

src/scss/components/_media-element.scss
80sgreek: It's a direct copy from INT CW, I can change it if you want but can we still aim to get this in today?

src/scss/components/_media-element.scss
80sgreek: It's a direct copy from INT CW, I can change it if you want but can we still aim to get this in today?

80sgreek: Actually, I completely agree even if it is a cut and paste :laughing:

80sgreek:
@tommaton My understanding of the 'o-' / 'c-' was that it was that 'o-' was for classes on native tags like '<button />' or '<a />'. If I'm wrong someone should probably explain it to me again as I'm adding it for other people's benefit, not mine :smile:

DamianMullins:
@tommaton @80sgreek I'll PM you the link to the CSS naming scheme (it's an internal link)

PR #64, opened 2018-02-28
v0.33.0 - Restaurant details wrapper (expressed as 'listings')

80sgreek:


src/scss/components/_cards.scss
```diff
@@ -15,7 +15,7 @@ $card-radius                    : 4px !default;
 
 @mixin card() {
 
-    .card {
+    .c-card {
```
ashleynolan: Not anything to change here, but will need to let the Payment and OrderWeb teams know that this class has changed when it goes in so they can update any relevant classes.

src/scss/components/_listings.scss
```diff
@@ -0,0 +1,65 @@
...
+                text-decoration: none;
+            }
+
+            &:hover {
+                text-decoration: none;
+
+                & .c-listing-item-title {
+                    text-decoration: underline;
+                }
+                & .o-btn--primary {
```
ashleynolan: It's better not to change other components within another component (so that there's no dependency between CSS components).

So this hook would be better declared as `c-listing-btn` and adding that class to the button in the listings.

src/scss/components/_listings.scss
```diff
@@ -0,0 +1,65 @@
...
+ * Listing
+ * ===================================
+ * Example – SERP List – Restaurant Information
+ *
+ *
+ */
+
+$listing-default-borderColor: $color-border;
+
+.c-listing {
```
ashleynolan: More a question for Damian for future work than a request for any changes on this PR, but would there ever be the possibility of being able to use the HTML sharing templates and pull this out into a full reusable HTML component?

src/scss/components/_listings.scss
```diff
@@ -0,0 +1,65 @@
+/**
+ * Listing
+ * ===================================
+ * Example – SERP List – Restaurant Information
+ *
+ *
+ */
+
+$listing-default-borderColor: $color-border;
```
ashleynolan: Can likely get rid of the default here I reckon?  Makes sense as just `$listing-borderColor`

ashleynolan:
Looks good – just a couple of comments 👍

src/scss/components/_cards.scss
80sgreek: Good point, will do.

80sgreek:
@ashleynolan thanks for the review, I'll get on it

src/scss/components/_listings.scss
DamianMullins: So, you mean we could have an `f-listings` npm module which would contain the SCSS and the templates @ashleynolan? I don't see why not, we just need to look a bit deeper into how we can compose templates together.

ashleynolan: Yeh – that's how I imagined it.

Maybe worth opening a JIRA, as it's not a massive priority right now, but would be cool to look into.  It's definitely a pattern that'll get re-used on other pages.

80sgreek:
Just fixing the build as lint errors are breaking it

PR #65, opened 2018-03-02
v0.34.0 - Ported over radio buttons and checkboxes from international 

tommaton:
Added new styling for form radio and checkboxes
Updated version of project

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 v0.34.0
 ------------------------------
-*March 5, 2018*
+*March 6, 2018*
 
 ### Added
 - Styling for radio buttons and checkboxes (crosses and ticks)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 v0.34.0
 ------------------------------
-*March 8, 2018*
+*March 2, 2018*
 
 ### Added
 - Styling for radio buttons and checkboxes (crosses and ticks)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 v0.34.0
 ------------------------------
-*March 6, 2018*
+*March 8, 2018*
 
 ### Added
 - Styling for radio buttons and checkboxes (crosses and ticks)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 v0.34.0
 ------------------------------
-*March 2, 2018*
+*March 5, 2018*
 
 ### Added
 - Styling for radio buttons and checkboxes (crosses and ticks)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,15 @@
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+v0.34.0
+------------------------------
+*March 8, 2018*
+
+### Added
+- Styling for radio buttons and checkboxes. Checkboxes having 2 styles
+-- Crosses
+-- Ticks
+
 v0.33.0
 ------------------------------
 *February 28, 2018*
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,9 +8,7 @@ v0.34.0
 *March 8, 2018*
 
 ### Added
-- Styling for radio buttons and checkboxes. Checkboxes having 2 styles
--- Crosses
--- Ticks
+- Styling for radio buttons and checkboxes (crosses and ticks)
 
 v0.33.0
 ------------------------------
--- a/src/scss/components/_form-elements.scss
+++ b/src/scss/components/_form-elements.scss
@@ -1,0 +1,241 @@
+/**
+ * Custom Checkboxes and radios
+ *
+ * --------------------------------
+ *
+ * Checkbox button example HTML
+ *
+ *    <label class="c-form-label" for="checkbox1">
+ *        <input class="c-form-input" type="checkbox" id="checkbox1" name="myCheckbox" value="A value"/>
+ *        <span class="c-form-indicator c-form-indicator--checkbox"></span>
+ *        My Checkbox Label Text
+ *    </label>
+ *
+ * -------------------------------
+ *
+ * Radio button example HTML
+ *
+ *    <label class="c-form-label" for="radio1">
+ *        <input class="c-form-input" type="radio" id="radio1" name="myRadioBtn" value="A value"/>
+ *        <span class="c-form-indicator c-form-indicator--radio"></span>
+ *        My Radio Button Label Text
+ *     </label>
+ *
+ * ================================= */
+ $input-bgColor                  : $white !default;
+ $input-bgColor-disabled         : $color-disabled;
+ $input-borderColor              : $color-border;
+ $input-color             : $blue !default;
+ $input-background-color  : $color-bg !default;
+ $input-hover-color       : $blue--dark !default;
+ $input-borderColor-hover : $blue !default;
+ $input-margin-left       : 0 !default;
+ $input-padding-left      : spacing(x4) !default;
+ $input-width--wide       : 20px !default;
+ $input-width             : 24px !default;
+ $input-height--wide      : 20px !default;
+ $input-height            : 24px !default;
+ $input-border-width      : 1px;
+
+
+ .c-form-label {
+     position: relative;
+     display: block;
+     cursor: pointer;
+     user-select: none;
+     padding: spacing();
+     padding-left: $input-padding-left;
+
+     &:hover .c-form-indicator {
+         border-color: $input-borderColor-hover;
+     }
+
+     &:hover .c-form-input:disabled ~ .c-form-indicator {
+         border-color: $input-borderColor;
+         cursor: not-allowed;
+     }
+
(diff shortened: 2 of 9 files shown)


CHANGELOG.md
```diff
@@ -3,6 +3,13 @@
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+v0.34.0
+------------------------------
+*March 8, 2018*
```
ashleynolan: Is this just a super pessimistic estimate? 🤣 ⌛️

src/scss/components/_form-elements.scss
```diff
@@ -0,0 +1,241 @@
...
+ $input-margin-left       : 0 !default;
+ $input-padding-left      : spacing(x4) !default;
+ $input-width--wide       : 20px !default;
+ $input-width             : 24px !default;
+ $input-height--wide      : 20px !default;
+ $input-height            : 24px !default;
+ $input-border-width      : 1px;
+
+
+ .c-form-label {
```
ashleynolan: As forms will use the `o-` syntax, would probably be good to follow with these as they are items within `form`.

Should hopefully just be a find and replace job (and moving the file to the objects folder)

src/scss/objects/_form-elements.scss
```diff
@@ -0,0 +1,241 @@
...
+ }
+
+ // Give inline elements some space to the right
+ .o-form-label--inline {
+     display: inline-block;
+     padding-right: 2em;
+ }
+
+     // The hidden input
+     .o-form-input {
```
ashleynolan: As this isn't called `control-input` I'd probably call it something a little more specific to checkboxes and radios, else it may clash with other text inputs when we get around to styling those.

Not sure what best to name it though.  Maybe `o-form-control` or something like that? (as that then represents both input types)

src/scss/objects/_form-elements.scss
```diff
@@ -0,0 +1,241 @@
...
+ $input-margin-left       : 0 !default;
+ $input-padding-left      : spacing(x4) !default;
+ $input-width--wide       : 20px !default;
+ $input-width             : 24px !default;
+ $input-height--wide      : 20px !default;
+ $input-height            : 24px !default;
+ $input-border-width      : 1px;
+
+
+ .o-form-label {
```
ashleynolan: Will this work for all form labels potentially?  Or is it specific to checkbox/radio labels.

Just want to avoid having to change the classname later if it's more specific

ashleynolan:
@tommaton Is this ready to review again? Will need a date update in the CHANGELOG too

tommaton:
@ashleynolan yep its ready to review again. I've also just pushed the changelog with the updated date.

src/scss/objects/_form-controls.scss
```diff
@@ -0,0 +1,240 @@
+/**
+ * Custom Checkboxes and radios
+ *
+ * --------------------------------
+ *
+ * Checkbox button example HTML
+ *
+ *    <label class="c-form-label" for="checkbox1">
```
DamianMullins: These docs need to be updated following the class name updates.

src/scss/objects/_form-controls.scss
```diff
@@ -0,0 +1,240 @@
...
+ * Radio button example HTML
+ *
+ *    <label class="c-form-label" for="radio1">
+ *        <input class="c-form-input" type="radio" id="radio1" name="myRadioBtn" value="A value"/>
+ *        <span class="c-form-indicator c-form-indicator--radio"></span>
+ *        My Radio Button Label Text
+ *     </label>
+ *
+ * ================================= */
+ $formControl-bgColor                  : $white !default;
```
DamianMullins: Any reason why some of these have been marked as `!default` and not others?

src/scss/objects/_form-controls.scss
```diff
@@ -0,0 +1,240 @@
...
+     }
+ }
+
+ // Give inline elements some space to the right
+ .o-formControl-label--inline {
+     display: inline-block;
+     padding-right: 2em;
+ }
+
+     // The hidden input
```
DamianMullins: Not sure what is meant by this comment? If it's a hidden input shouldn't the selector be named to describe what it is doing? Could also be added as a modifier to `o-formControl-input`?

src/scss/objects/_form-controls.scss
```diff
@@ -0,0 +1,240 @@
...
+             border-radius: 50%;
+             width: $formControl-width - 2;
+             height: $formControl-width - 2;
+             position: absolute;
+             top: -5px;
+             left: -5px;
+
+             @include media('>=mid') {
+                 width: $formControl-width--wide - 2;
+                 height: $formControl-width--wide - 2;
```
kevinrodrigues: `$formControl-height--wide`

src/scss/objects/_form-controls.scss
```diff
@@ -0,0 +1,240 @@
...
+ $formControl-bgColor                  : $white !default;
+ $formControl-bgColor-disabled         : $color-disabled;
+ $formControl-borderColor              : $color-border;
+ $formControl-color             : $blue !default;
+ $formControl-background-color  : $color-bg !default;
+ $formControl-hover-color       : $blue--dark !default;
+ $formControl-borderColor-hover : $blue !default;
+ $formControl-margin-left       : 0 !default;
+ $formControl-padding-left      : spacing(x4) !default;
+ $formControl-width--wide       : 20px !default;
```
kevinrodrigues: I feel this should be 24px and then the `$formControl-width` should be 20px? Or am I misunderstanding the `--wide` bit?

src/scss/objects/_form-controls.scss
```diff
@@ -0,0 +1,240 @@
...
+
+         &:focus {
+             & + .o-formControl-indicator { @extend %u-elementFocus; }
+             & ~ .o-formControl-text { @extend %u-elementFocus; }
+         }
+
+         .no-csstransforms & {
+             position: relative !important;
+             opacity: 1;
+             z-index: 1;
```
kevinrodrigues: We should make use of the `$zIndex` map property here and anywhere else z-index is used, to keep them all consistent and in one place.

src/scss/objects/_form-controls.scss
```diff
@@ -0,0 +1,240 @@
...
+         height: $formControl-height;
+         border: $formControl-border-width solid $grey--lightest;
+         background-color: $formControl-bgColor;
+         vertical-align: middle;
+         margin-left: $formControl-margin-left;
+         margin-top: 0;
+         box-shadow: 0 0 0 2px transparent, 0 0 0 0 transparent; // Used to animate from when element is in :focus
+
+         .no-csstransforms & {
+             display: none;
```
kevinrodrigues: More of a question, does this element not appear if css transforms are not supported? Could it not fallback to no animation?

src/scss/objects/_form-controls.scss
tommaton: Perhaps not the best description. Its to hide the default radio or checkbox so you only see the label and the designed radio/checkbox

src/scss/objects/_form-controls.scss
tommaton: I've added them in myself and forgot to add them on.

src/scss/objects/_form-controls.scss
tommaton: @kevinrodrigues do you have an example of this as I've just ported this over from International.

kevinrodrigues: Something like `$zIndex(low)` should do it.

src/scss/objects/_form-controls.scss
ashleynolan: Might be better to change this to be `$formControl-diameter--wide` as it's a square, so the same value will be used for width/height

src/scss/objects/_form-controls.scss
ashleynolan: The wide modifier here is for a media query.  Might be worth adding a comment saying that though as agree could be confusing

kevinrodrigues: Ahh I see.

src/scss/objects/_form-controls.scss
ashleynolan: @tommaton Just for reference, [this is the map](https://github.com/justeat/fozzie/blob/master/src/scss/settings/_variables.scss#L113-L117) @kevinrodrigues is referring to.  If you need it to be specifically 1, then you can add a value to that map and reference that

src/scss/objects/_form-controls.scss
ashleynolan: We might be able to get rid of this altogether now – I think it was a bit of a hack rather than anything else, but now our support levels are up to IE11, we probably don't need this anymore (and any other `no-csstransforms` references)

src/scss/objects/_form-controls.scss
tommaton: Cheers guys, I'll make that update

src/scss/objects/_form-controls.scss
tommaton: @ashleynolan `$formControl-diameter--wide` to me would be referencing a circle (radio button) so for me this would be incorrect as the width is also for a square (checkbox).

ashleynolan: `length`?

src/scss/objects/_form-controls.scss
```diff
@@ -0,0 +1,230 @@
...
+ .o-formControl-label--inline {
+     display: inline-block;
+     padding-right: 2em;
+ }
+
+     // Hide the input element from view
+     .o-formControl-input {
+         position: absolute;
+         opacity: 0;
+         z-index: -1; // Put the input behind the label so it doesn't overlay text
```
kevinrodrigues: It would be really useful to have this added to [here](https://github.com/justeat/fozzie/blob/master/src/scss/settings/_variables.scss#L113) maybe as `lowest: -1` I can see it being useful in other areas.

CHANGELOG.md
tommaton: Perhaps I should of left it now, lol

ashleynolan:
@DamianMullins @kevinrodrigues Can you guys comment/approve now the changes have been made?

PR #69, opened 2018-03-19
v0.39.0 - selectable star ratings and non selectable star ratings

tommaton:
Ported over ratings for new selectable CSS stars without JS
Ported over existing ratings with couple more extenders
Tidied up existing CSS

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ v0.39.0
 
 ### Added
 - Selectable star ratings
-- Non selectable start ratings
+- Non selectable star ratings
 - Mixin to generate fill classes for ratings
 
 v0.38.0
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,10 +5,18 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 v0.38.0
 ------------------------------
+<<<<<<< HEAD
 *April 10, 2018*
 
 ### Changed
 - Font size variable map updated.
+=======
+*March 19, 2018*
+
+### Added
+- Selectable star ratings
+- Non selectable start ratings
+>>>>>>> Development of Selectable star ratings and non selectable star ratings
 
 v0.37.0
 ------------------------------
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,20 +3,21 @@
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
-v0.38.0
+v0.39.0
 ------------------------------
-<<<<<<< HEAD
-*April 10, 2018*
-
-### Changed
-- Font size variable map updated.
-=======
-*March 19, 2018*
+*April 17, 2018*
 
 ### Added
 - Selectable star ratings
 - Non selectable start ratings
->>>>>>> Development of Selectable star ratings and non selectable star ratings
+- Mixin to generate fill classes for ratings
+
+v0.38.0
+------------------------------
+*April 10, 2018*
+
+### Changed
+- Font size variable map updated.
 
 v0.37.0
 ------------------------------
--- a/src/scss/components/_ratings.scss
+++ b/src/scss/components/_ratings.scss
@@ -12,22 +12,22 @@ $star-size-large: 42px;
 }
 
 .c-ratings--medium {
-   height: $star-size-medium;
-   width: $star-size-medium * $star-count;
+    height: $star-size-medium;
+    width: $star-size-medium * $star-count;
 }
 
 .c-ratings--large {
-   height: $star-size-large;
-   width: $star-size-large * $star-count;
+    height: $star-size-large;
+    width: $star-size-large * $star-count;
 }
 
 .c-ratings--fill {
-   background: inline('#{$icon-path}/stars/star--filled.svg') repeat-x;
-   background-size: contain;
-   display: inline-block;
-   height: 100%;
-   width: 100%;
-   @include rating-fill($star-count);
+    background: inline('#{$icon-path}/stars/star--filled.svg') repeat-x;
+    background-size: contain;
+    display: inline-block;
+    height: 100%;
+    width: 100%;
+    @include rating-fill($star-count);
 }
 
 /*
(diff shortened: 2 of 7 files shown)


src/scss/components/_ratings.scss
```diff
@@ -0,0 +1,157 @@
...
+*
+* Non-Selectable star rating
+*
+*   <div class="c-ratings ">
+*      <span class="c-ratings--fill c-ratings--fill-18"></span>
+*   </div>
+*
+* ================================= */
+
+$star-filled: "data:image/svg+xml;charset=utf-8,%3Csvg xmlns='http://www.w3.org/2000/svg' width='13' height='13' viewBox='0 0 13 13'%3E%3Cpath fill='%23D50525' d='M6.5 10.8l-4 2.2.8-4.7L0 5l4.5-.7 2-4.3 2 4.3L13 5 9.8 8.3l.7 4.7z'/%3E%3C/svg%3E";
```
ashleynolan: Icons should either use the SVG map or the `inline()` function and the icons themselves should live in [`f-icons`](https://github.com/justeat/f-icons)

src/scss/components/_ratings.scss
```diff
@@ -0,0 +1,63 @@
...
+.c-ratings {
+    background: inline('#{$icon-path}/stars/star--empty.svg') repeat-x;
+    background-size: contain;
+    display: inline-block;
+    height: $star-size-small;
+    width: $star-size-small * $star-count;
+}
+
+.c-ratings--medium {
+   height: $star-size-medium;
```
kevinrodrigues: I think the indenting is out from here and the next few blocks.

src/scss/components/_ratings.scss
```diff
@@ -0,0 +1,63 @@
...
+           background-image: inline('#{$icon-path}/stars/star--filled.svg');
+        }
+   }
+}
+
+.c-icon--star {
+    display: block;
+    cursor: pointer;
+    background-image: inline('#{$icon-path}/stars/star--empty.svg');
+    text-indent: -9999px;
```
kevinrodrigues: Would using `@extend %u-hideText;` work better here?

src/scss/tools/mixins/_ratings.scss
```diff
@@ -0,0 +1,33 @@
...
+                    &.c-ratings--fill-#{$i}#{$ratingIteration} {
+                        width: (($ratingIteration * 0.1) + $i) / $maxStars * 100%;
+                    }
+                }
+
+            }
+        }
+        @else {
+            &.c-ratings--fill-#{$i}#{0} {
+                width: $i / $maxStars * 100%;
```
kevinrodrigues: If someone passes maxStars = 0 then I think width would be NaN? Not sure this would actually be an issue use case wise of the mixin?

src/scss/tools/mixins/_ratings.scss
```diff
@@ -0,0 +1,37 @@
...
+
+$rating-iterations: (0, 2, 5, 6, 8);
+
+@mixin rating-fill($maxStars: 6) {
+    @if $maxStars > 0 {
+        @for $i from 0 to $maxStars + 1 {
+            @if $i < $maxStars {
+                @each $ratingIteration in $rating-iterations {
+                    @if $i == 0 and $ratingIteration == 0 {
+                        &.c-ratings--fill-00 {
```
DamianMullins: Stylelint is failing the Travis build as the nesting on this line and line 21 exceeds the maximum indentation level. I think it's OK to wrap this mixin with `/* stylelint-disable max-nesting-depth */` and `/* stylelint-enable max-nesting-depth */`?

src/scss/tools/mixins/_ratings.scss
```diff
@@ -0,0 +1,37 @@
+//
+// ==========================================================================
+// Ratings variant mixin
+// Used in alert modifier classes to colorize the alert
+// ==========================================================================
```
ashleynolan: Would be good to include an example of the classes produced by this mixin.

Looks to be something like `.c-ratings--fill-02`, `.c-ratings--fill-05` etc, but good to know for certain if anyone needs to edit it down the line.

Also, would be good to explain why we are creating these classes using the mixin - which is to allow us to control the width of the star ratings using classnames?

CHANGELOG.md
```diff
@@ -3,6 +3,15 @@
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+v0.38.0
+------------------------------
+*April 6, 2018*
+
+### Added
+- Selectable star ratings
+- Non selectable start ratings
```
ashleynolan: Typo `star`

src/scss/components/_ratings.scss
```diff
@@ -0,0 +1,63 @@
...
+    justify-content: flex-end;
+}
+
+.c-ratingSelectable-input {
+    display: none;
+
+    &:checked {
+        & + label,
+        & ~ input + label {
+           background-image: inline('#{$icon-path}/stars/star--filled.svg');
```
ashleynolan: Would make sense to make a silent extends class for the icons.  At the moment, if the icon name for filled stars changed, it would need to be changed in 3 places.  Would be better if it was one place 🔁

src/scss/components/_ratings.scss
```diff
@@ -0,0 +1,63 @@
...
+        }
+   }
+}
+
+.c-icon--star {
+    display: block;
+    cursor: pointer;
+    background-image: inline('#{$icon-path}/stars/star--empty.svg');
```
ashleynolan: Same as above - would be good to use a silent extends class for this to help with maintenance.

src/scss/components/_ratings.scss
ashleynolan: It probably makes sense to be a part of `f-icons` as I would think it makes most sense to call the classes `%c-icon--star--fill` and ``%c-icon--star--empty`.

src/scss/components/_ratings.scss
```diff
@@ -0,0 +1,65 @@
+$star-count: 6;
+$star-size-small: 14px;
+$star-size-medium: 28px;
+$star-size-large: 42px;
+
+.c-ratings {
```
ashleynolan: Think lost my comment from before - would be good to keep this singular as it's only one rating.

So `c-rating` rather than `c-ratings`.  Otherwise looks good and happy to approve

PR #72, opened 2018-04-10
v0.38.0 - Updated font size map.

kevinrodrigues:


src/scss/settings/_variables.scss
```diff
@@ -38,12 +38,16 @@ $type: (
...
-        default: (28, 36),
+        default: (32, 32),
         narrow: (18, 24)
     ),
     jumbo--scaleUp: (
-        default: (64, 72),
+        default: (44, 48),
         narrow: (28, 32)
+    ),
+    mid--podTitle: (
```
ashleynolan: Should name this more generically as it may one day be used for things that aren't podTitles :)

src/scss/settings/_variables.scss
```diff
@@ -21,15 +21,15 @@ $type: (
...
-        default: (20, 28),
-        narrow: (18, 24)
+        default: (24, 28),
+        narrow: (20, 24)
     ),
     jumbo: ( // h1
-        default: (24, 32),
-        narrow: (20, 24)
+        default: (30, 40),
+        narrow: (24, 24)
```
ashleynolan: Worth testing all of these new combinations of font-sizes and line-height's.

Just tried testing this one myself - having the same font-size and line-height reads terribly when the text does wrap onto 2 lines, which is pretty likely on mobile.  Would be worth asking Bruno what line-height this should be set to instead as looks like an oversight

src/scss/settings/_variables.scss
```diff
@@ -38,12 +38,16 @@ $type: (
         narrow: (16, 21)
     ),
     large--scaleUp: (
-        default: (28, 36),
+        default: (32, 32),
```
ashleynolan: Same here - can't imagine this is intentional as the space between lines of text is very shallow

kevinrodrigues:
@ashleynolan Caught up with Bruno and got him to agree on removing the podTitle value as it was similar to the existing h2 styles we already have. Some of the other line height values weren't correct in the Zeplin file so he's update the file now and I've applied the new changes here. :)

ashleynolan:
@kevinrodrigues Cool - looks much better now.  If someone else can review as well happy to merge.

bulutcy:
I think this was supposed to be a breaking change, isn't it?
It doesn't break sass, but it breaks the design. I had to change the font size variables we use.

kevinrodrigues:
@bulutcy I'm not sure it's really a breaking change (functionally it works fine) but if it's breaking the design we should show the designers because they've requested these changes, if we change font variables to fix design issues we will end up with multiple different sizes to cater for fixes which is what we're trying to avoid here :) best to chat to Bruno Molena (he's the UI designer on this)

bulutcy:
Ok, I understand. If it's requested by the designer, it should be ok to use new font sizes.

ashleynolan:
As Kevin said – best thing to do would be to ask Bruno to review the font-sizes on the Order sections if any of it looks slightly off.  It might just be that you need to adjust some of the relative sizes from h2 > h3 size or something like that, but best to check with them about this as they can help advise.

PR #74, opened 2018-04-20
v0.41.0 - Fixed button font size issue.

kevinrodrigues:


src/scss/settings/_variables.scss
```diff
@@ -20,6 +20,9 @@ $type: (
     base: (  // body copy
         default: (14, 18)
     ),
+    mid--scaleDown: (
```
ashleynolan: Would it be better to name this as a `base--scaleUp `, just as it relates closer to the base sizing than `mid`?

PR #82, opened 2018-06-15
v0.45.0 - Updated font size.

kevinrodrigues:


CHANGELOG.md
```diff
@@ -3,6 +3,14 @@
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+v0.45.0
+------------------------------
+*June 15, 2018*
```
ashleynolan: Bump 😄

src/scss/settings/_variables.scss
```diff
@@ -41,8 +41,8 @@ $type: (
         narrow: (16, 20)
     ),
     large--scaleUp: (
-        default: (32, 40),
-        narrow: (18, 24)
+        default: (20, 24),
```
ashleynolan: The values no longer fit the name of this one.  Large is `24/28` so the name should change to `large--scaleDown` if it's there to transform that size down a little

ashleynolan: Also - any idea what currently uses this variable?

kevinrodrigues: It's used for the homepage strapline in the new hero slingshot work. Currently only in UK so I wanted to update it so GlobalWeb is inline. I'll update the name shortly.

src/scss/settings/_variables.scss
```diff
@@ -40,9 +40,9 @@ $type: (
         default: (22, 28),
         narrow: (16, 20)
     ),
-    large--scaleUp: (
-        default: (32, 40),
-        narrow: (18, 24)
+    large--scaleDown: (
```
xander-marjoram: Is this worth mentioning the in changelog?

kevinrodrigues: Yup updated.

PR #94, opened 2018-08-01
v0.53.0 – Addition of form toggle disabled state and touch area size

sonnyprince:
_Addition of form toggle disabled state and touch area size._

 - height unit on fullScreenOverlay
 - addition of .o-formToggle--disabled state styles
 - addition of form toggle inline spacing to give 44px touch area
 - addition of cursor pointer of form toggle hover

screen shot 2018-08-01 at 15 03 08

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 v0.53.0
 ------------------------------
-*August 1, 2018*
+*August 2, 2018*
 
 ### Added
  - addition of .o-formToggle--disabled state styles
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,10 +8,12 @@ v0.53.0
 *August 1, 2018*
 
 ### Changed
- - height unit on fullScreenOverlay
  - addition of .o-formToggle--disabled state styles
  - addition of form toggle inline spacing to give 44px touch area
  - addition of cursor pointer of form toggle hover
+
+### Fixed
+ - height unit on fullScreenOverlay from vh to % as Android didnt take into account browser bars making 100vh larger than 100%
 
 
 v0.52.0
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@ v0.53.0
 ------------------------------
 *August 1, 2018*
 
-### Changed
+### Added
  - addition of .o-formToggle--disabled state styles
  - addition of form toggle inline spacing to give 44px touch area
  - addition of cursor pointer of form toggle hover
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,8 @@ v0.53.0
 
 ### Changed
  - height unit on fullScreenOverlay
+ - addition of .o-formToggle--disabled state styles
+ - addition of form toggle inline spacing to give 44px touch area
 
 
 v0.52.0
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ v0.53.0
  - height unit on fullScreenOverlay
  - addition of .o-formToggle--disabled state styles
  - addition of form toggle inline spacing to give 44px touch area
+ - addition of cursor pointer of form toggle hover
 
 
 v0.52.0
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 v0.53.0
 ------------------------------
-*July 31, 2018*
+*August 1, 2018*
 
 ### Changed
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -3,6 +3,17 @@
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html)
 
+v0.53.0
+------------------------------
+*July 31, 2018*
```
xander-marjoram: Trivial, but can you bump this to today's date? Otherwise LGTM

CHANGELOG.md
```diff
@@ -3,6 +3,17 @@
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html)
 
+v0.53.0
+------------------------------
+*August 1, 2018*
+
+### Changed
+ - height unit on fullScreenOverlay
+ - addition of .o-formToggle--disabled state styles
```
ashleynolan: These should be put under `## Added` title, rather than changed

package.json
```diff
@@ -2,7 +2,7 @@
   "name": "@justeat/fozzie",
   "title": "Fozzie – Just Eat UI Web Framework",
   "description": "UI Web Framework for the Just Eat Global Platform",
-  "version": "0.51.3",
+  "version": "0.53.0",
```
ashleynolan: Is there a v52?  If not it's probably worth updating the changelog so that the last version was `v0.51.3` and this one is then `v0.52.0`

CHANGELOG.md
```diff
@@ -3,6 +3,17 @@
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html)
 
+v0.53.0
+------------------------------
+*August 1, 2018*
+
+### Changed
+ - height unit on fullScreenOverlay
```
ashleynolan: Always good to explain why – was this just a bug?  If so, this should be under `fixed` heading

src/scss/objects/_form-toggle.scss
```diff
@@ -11,10 +11,12 @@ $formToggle-border-color-checked    : $grey--lighter;
 $formToggle-border-width            : 1px;
 $formToggle-border-radius           : 8px;
 $formToggle-tick-background         : $green;
+$formToggle-tick-hover-background   : $grey--lighter;
```
ashleynolan: Wonder if this (and the variable above) should be `formToggle-icon-hover-background` so that if the icon changes in future the variable name doesn't need to be updated.

src/scss/objects/_form-toggle.scss
```diff
@@ -44,6 +47,10 @@ $formToggle-button-background       : $blue--offWhite;
         fill: $formToggle-button-color;
     }
 
+    .o-formToggle--inline {
```
ashleynolan: Always good to comment modifiers to say what they do when they are applied.  This one looks like it's adding spacing, but the name is a bit ambiguous as it implies it would make the formToggle inline?

src/scss/objects/_form-toggle.scss
```diff
@@ -164,3 +182,8 @@ $formToggle-button-background       : $blue--offWhite;
         right: spacing();
         transform: translateY(-50%);
     }
+
+    .o-formToggle--disabled {
```
ashleynolan: Is this class just used to add disabled styling to the parent?

package.json
sonnyprince: the previous release was v52 but typed in the wrong number. Do we need to release a minor with a 52 version bump before this?

src/scss/objects/_form-toggle.scss
sonnyprince: that's correct :)

package.json
ashleynolan: I would yeh, just to keep the changelog correct with the releases on NPM, as there isn't a `v0.52.0` on there atm

CHANGELOG.md
```diff
@@ -3,6 +3,19 @@
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html)
 
+v0.53.0
+------------------------------
+*August 1, 2018*
+
+### Changed
```
ashleynolan: Looks like these should be under `## Added` title, rather than changed, if they are additions

package.json
sonnyprince: PR for you 😁 
https://github.com/justeat/fozzie/pull/95

ashleynolan: I've merged that one – if you can release if this one should be good to go then too

sonnyprince:
@ashleynolan @xander-marjoram changes complete 😬

PR #95, opened 2018-08-01
#trivial - package.json version bump

sonnyprince:
__*version bump in package.json*__


ashleynolan:
@sonnyprince I gacve you some wrong info – to skip the build you need to use `#trivial` 😊  I've updated the title…

PR #96, opened 2018-08-02
v0.53.1 - Fixed stars look like selectable when the input is disabled.

bulutcy:
Fixed stars look like selectable when the input is disabled.
If the input is disabled, stars won't lit on hover.


xander-marjoram:
@bulutcy Do you have any screenshots?

bulutcy:
@xander-marjoram there is no visual change to show in a screenshot really.

PR #98, opened 2018-08-09
v0.54.0 – Form reset module

sonnyprince:
_Fozzie form reset module (JS only)._

## Browsers Tested

- [x] Chrome
- [x] Edge
- [x] Internet Explorer 11

### List any other browsers that this PR has been tested in:

- [View the Browser Support Checklist](http://fozzie.just-eat.com/documentation/general/browser-support)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,13 @@
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html)
 
+v0.54.0
+------------------------------
+*August 9, 2018*
+
+### Added
+ - addition of form reset javascript module and associated unit tests
+
 v0.53.1
 ------------------------------
 *August 2, 2018*
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,10 +5,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 v0.54.0
 ------------------------------
-*August 9, 2018*
+*August 13, 2018*
 
 ### Added
- - addition of form reset javascript module and associated unit tests
+ - Addition of form reset javascript module and associated unit tests
 
 v0.53.1
 ------------------------------
--- a/src/js/index.js
+++ b/src/js/index.js
@@ -1,7 +1,9 @@
 // All helper functions will be imported here, so that they can all be exported within one object.
 import { getBreakpoints, getCurrentScreenWidth } from './modules/breakpointHelper';
+import formReset from './modules/formReset';
 
 export {
+    formReset,
     getBreakpoints,
     getCurrentScreenWidth
 };
--- a/src/js/modules/formReset/test/index.test.js
+++ b/src/js/modules/formReset/test/index.test.js
@@ -7,7 +7,7 @@ describe('module init', () => {
         expect(typeof init).toBe('function');
     });
 
-    it('will not error if no [data-js-reset] is available', () => {
+    it('will check if a [data-js-reset] is available', () => {
         // Arrange
         TestUtils.setBodyHtml('');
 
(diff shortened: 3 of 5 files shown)


src/js/index.js
```diff
@@ -1,7 +1,9 @@
 // All helper functions will be imported here, so that they can all be exported within one object.
 import { getBreakpoints, getCurrentScreenWidth } from './modules/breakpointHelper';
+import formReset from './modules/formReset';
 
 export {
+    formReset,
```
ashleynolan: Would be better if we could make this optional as an export, as it won't be needed on all pages on the site.

Not sure what the best way of doing this would be while enabling easy access to the module from within Fozzie (so people don't have annoying JS absolute paths into module directories).  Any thoughts @DamianMullins @kevinrodrigues @simonsandersje ?

Kinda thinking it would be a little like lodash helpers, where you can import the ones you need as you need them – but equally not sure having a repo for every helper would be the best idea from a maintenance point of view (unless we have a separate folder in fozzie which contains these helpers and we release from these subfolders? Thinking how airbnb do it with the different eslint flavours – they're in one repo, with different package.json's within that repo)

ashleynolan:
Actual code looks good – would just be good to resolve (at least so we can open a JIRA and the work can get added to a backlog) how we store and pull in submodules that aren't needed by all microsites.

src/js/modules/formReset/test/index.test.js
```diff
@@ -0,0 +1,149 @@
...
+
+        // Act
+        init();
+        const button = $('[data-js-reset]');
+
+        // Assert
+        expect(button.length).toBe(0);
+    });
+
+    it('will uncheck a radio buttons from from button trigger bound on init()', () => {
```
kevinrodrigues: Not sure this one reads right.

src/js/modules/formReset/test/index.test.js
```diff
@@ -0,0 +1,149 @@
...
+        init();
+        const [button] = document.querySelectorAll('[data-js-reset]');
+        TestUtils.click(button);
+
+        // Assert
+        const [radio] = document.querySelectorAll('[data-js-reset-input]');
+        expect(radio.checked).toBe(false);
+    });
+
+    it('will clear multiple form elements from button trigger bound on init()', () => {
```
kevinrodrigues: Could you put this in a describe block then for each expect below scope it to one single `IT`.


```
describe('When multiple form elements are cleared', () => {
    it('should uncheck a radio form element from button trigger bound on init()', () => {
        TestUtils.setBodyHtml(`
                    <a data-js-reset="testGroup"></a>
                    <input type="radio" data-js-reset-input="testGroup" checked/>
                `);
        const [radio] = document.querySelectorAll('[data-js-reset-input]');
        expect(radio.checked).toBe(false);
    })

    ...
    ...
});
```

src/js/modules/formReset/test/index.test.js
```diff
@@ -0,0 +1,149 @@
...
+
+        // Act
+        resetInputs(event);
+
+        // Assert
+        const [input] = document.querySelectorAll('[data-js-reset-input]');
+        expect(input.value.length).toBe(0);
+    });
+
+    it('wont error if reset form button is clicked but no elements need resetting', () => {
```
kevinrodrigues: `Won\'t`

Is this test testing for errors or states? Does the radio button and checkbox need the `checked` attribute removing?

Also worth using a describe block here.

src/js/index.js
kevinrodrigues: I think that's the only way, to have a single repo of helpers in individual folders (containing the index.js, package.json & readme etc) under our fozzie namespace then we can release it all in one go and import them when we need them.

CHANGELOG.md
```diff
@@ -3,6 +3,13 @@
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html)
 
+v0.54.0
+------------------------------
+*August 9, 2018*
```
xander-marjoram: Date bump please

src/js/modules/formReset/test/index.test.js
```diff
@@ -0,0 +1,167 @@
...
+            init();
+            const [button] = document.querySelectorAll('[data-js-reset]');
+            TestUtils.click(button);
+    
+            // Assert
+            const [checkbox] = document.querySelectorAll('[data-js-reset-input]');
+            expect(checkbox.checked).toBe(false);
+        });
+
+        it('should uncheck a checkbox form element from button trigger bound on init', () => {
```
kevinrodrigues: `it('should clear the 'value' of a text input...'`

PR #104, opened 2018-08-15
v0.59.1 - Content Title Action hover state

sonnyprince:
__*Content Title Action hover state*__

- change text decoration of content header from none to underlined on hover

Before:
screen shot 2018-08-15 at 15 21 40

After:
screen shot 2018-08-17 at 11 19 08

---

## UI Review Checks

## Browsers Tested

- [x] Chrome
- [x] Mobile

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,13 @@
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html)
 
+v0.59.1
+------------------------------
+*August 22, 2018*
+
+### Changed
+ - Change content header action hover state from no decoration normal and underline when hovered.
+
 v0.59.0
 ------------------------------
 *August 22, 2018*
--- a/src/scss/components/_content-title.scss
+++ b/src/scss/components/_content-title.scss
@@ -26,4 +26,9 @@
 
     .c-contentTitle-action {
         font-weight: $font-weight-bold;
+        text-decoration: none;
+
+        &:hover {
+            text-decoration: underline;
+        }
     }
--- a/src/scss/components/_content-title.scss
+++ b/src/scss/components/_content-title.scss
@@ -28,7 +28,8 @@
         font-weight: $font-weight-bold;
         text-decoration: none;
 
-        &:hover {
+        &:hover,
+        &:focus {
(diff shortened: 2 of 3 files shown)


src/scss/components/_content-title.scss
```diff
@@ -26,4 +26,9 @@
 
     .c-contentTitle-action {
         font-weight: $font-weight-bold;
+        text-decoration: none;
+
+        &:hover {
```
kevinrodrigues: :focus?

sonnyprince: updated 😃

src/scss/components/_content-title.scss
```diff
@@ -26,4 +26,10 @@
 
     .c-contentTitle-action {
         font-weight: $font-weight-bold;
+        text-decoration: none;
```
ashleynolan: Accessibility Q – is this action obvious enough for anyone who is colourblind if you get rid of the underline?

I know we've pushed back on things like this in the past (like on the [registration pages](https://www.just-eat.es/account/register) in the t&cs text, where it isn't obvious otherwise that it's a clickable link)

sonnyprince: V good point - I'll raise with Design.

piet-je: @ashleynolan  The context and positioning (right aligned to the title) and Medium weight of the fon (above screenshots are in the incorrect font) should help with this. The styling for this CTA link follows our default CTA link styling that's in use across the product. If there is a concern this styling is not accessible, it'd possibly need a full site review of some kind.

ashleynolan: @piet-je Yeh – it'd be good if we could look at this potentially in the future.  I'd be interested to see how it performed in accessibility tests with real users (not sure if that's something we do across our designs – I know we've done it in the past).

xander-marjoram:
@sonnyprince You'll need to merge in latest master and update the version number now that 0.55.0 has gone in, as well as the date in the `package.json`.

CHANGELOG.md
```diff
@@ -3,6 +3,13 @@
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html)
 
+v0.54.1
```
ashleynolan: Will need to version bump this as another PR has gone in since

PR #105, opened 2018-08-16
v0.54.0 - addition of has-no-match styling to c-listing component

sonnyprince:
__*addition of has-no-match styling to c-listing component*__

## UI Review Checks

Before:
screen shot 2018-08-16 at 14 55 53

After:
screen shot 2018-08-16 at 14 55 59

## Browsers Tested

- [x] Chrome
- [x] Edge
- [x] Firefox
- [x] Internet Explorer 11
- [x] Mobile

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,13 @@
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html)
 
+v0.54.0
+------------------------------
+*August 16, 2018*
+
+### Added
+ - Addition of `has-no-match` class and styling to `.c-listing` item. (removes border).
+
 v0.53.4
 ------------------------------
 *August 15, 2018*
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ v0.54.0
 *August 16, 2018*
 
 ### Added
- - Addition of `has-no-match` class and styling to `.c-listing` item. (removes border).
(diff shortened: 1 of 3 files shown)


src/scss/components/_listings.scss
```diff
@@ -34,6 +34,9 @@ $listing-borderRadius       : 4px;
         }
     }
 
+    &.has-no-match {
```
ashleynolan: Should make this helper relate to what it is, rather than what it does.  Otherwise it'll be a bit strange in other listings where `no-match` doesn't mean much.

Either that, or I'd consider making this a full on utility which (`u-stripBorder`?) which can be added potentially

sonnyprince: what would you call the class name? `has-child-listings`, `has-children`?
Not sure I could use a utility - I need an `is` or `has` class for the JS side of things :)

ashleynolan: Possibly a better naming actually would just be `has-noResults` as for a listing component that would work?

sonnyprince: consider it done!

CHANGELOG.md
```diff
@@ -3,6 +3,13 @@
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html)
 
+v0.54.0
+------------------------------
+*August 16, 2018*
+
+### Added
+ - Addition of `has-no-match` class and styling to `.c-listing` item. (removes border).
```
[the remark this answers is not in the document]
sonnyprince: done 👍 😄

PR #110, opened 2018-08-21
v0.58.0 — Updates to badge component.

DamianMullins:
### Added
- New badge component modifiers

### Fixed
- Added prefix to badge component classes.

### DIFF
--- a/src/scss/components/_badges.scss
+++ b/src/scss/components/_badges.scss
@@ -10,6 +10,7 @@
  */
 
 $badge-bgColor      : $grey--offWhite !default;
+$badge-lightBgColor : $white !default;
 $badge-infoColor    : $green !default;
 $badge-warningColor : $orange !default;
 $badge-alertColor   : $red !default;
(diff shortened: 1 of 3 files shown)


src/scss/components/_badges.scss
```diff
@@ -18,25 +19,37 @@ $badge-radius       : 14px !default;
...
 @mixin badge() {
 
-    .badge {
+    .c-badge {
         background-color: $badge-bgColor;
         padding: $badge-padding;
     }
 
-        .badge--rounded {
+        .c-badge--rounded {
```
oanuna: things seem indented by one tab below here

DamianMullins: I thought that was the convention for modifiers etc? I was [following the style used in the `cards` component](https://github.com/justeat/fozzie/blob/master/src/scss/components/_cards.scss).

oanuna: i think if it was nested like so it would make sense:

```scss
.c-card {
  &--noPad {
    ...
  }
  &--noPadVertical {
    ...
  }
}
```

but as it all sits inside a single level i would keep them tabbed in line

ashleynolan: That's the current convention we use across other components.  Prob best to leave it like this and then if we want to change it do so all in one go (can discuss if we think it's wrong atm).

DamianMullins:
So, are we happy to get this merged as is and have a conversation around structure soon?

PR #111, opened 2018-08-22
v0.59.0 - Addition of is sticky utility

sonnyprince:
_Addition is `is-sticky` util class._

That is all :) 

(Version is +1 in the future assuming @DamianMullins v0.58.0 will go in first)


DamianMullins:
v0.58.0 (#110) has gone in now @sonnyprince

PR #115, opened 2018-08-23
v0.61.0 — Added new utility helpers.

DamianMullins:
### Added
- Centre align utility.
- No underline utility.

### Removed
- Media element centered element.

### DIFF
--- a/src/scss/base/_layout.scss
+++ b/src/scss/base/_layout.scss
@@ -45,6 +45,15 @@ body {
         padding-top: spacing(x4);
     }
 }
+
+    // uses flexbox to centre align content both vertically and horizontally
+    .l-content--centred, // British or US English spellings..
+    .l-content--centered {
+        align-items: center;
+        display: flex;
+        justify-content: center;
+    }
+
     .l-content-header {
         margin: 0 0 spacing(x2);
 
--- a/src/scss/base/_layout.scss
+++ b/src/scss/base/_layout.scss
@@ -46,14 +46,6 @@ body {
     }
 }
 
-    // uses flexbox to centre align content both vertically and horizontally
-    .l-content--centred, // British or US English spellings..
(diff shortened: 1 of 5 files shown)


src/scss/trumps/_utilities.scss
```diff
@@ -112,6 +120,10 @@
     padding-left: spacing(x4);
 }
 
+.u-noUnderline {
+    text-decoration: none;
```
ashleynolan: It's probably worth this helper having `:hover` and `:focus` state which applies the underline

DamianMullins: Done.

src/scss/trumps/_utilities.scss
```diff
@@ -112,6 +120,16 @@
...
+
+    &:hover,
+    &:active,
+    &:focus {
+        text-decoration: none;
```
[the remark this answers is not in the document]
DamianMullins: 🤦‍♂️ Done

DamianMullins:
@ashleynolan I've just noticed the [`l-inlined` class](https://github.com/justeat/fozzie/blob/master/src/scss/base/_layout.scss#L80-L87) which is very similar to the `u-centered` class, should we move the new class to sit alongside that one do you think?

ashleynolan:
@DamianMullins Yep 👀

src/scss/base/_layout.scss
```diff
@@ -45,6 +45,15 @@ body {
         padding-top: spacing(x4);
     }
 }
+
+    // uses flexbox to centre align content both vertically and horizontally
+    .l-content--centred, // British or US English spellings..
```
ashleynolan: Guessing this would just be `l-centered` as otherwise it implies it would always be used with the `l-content` class as a modifier?

PR #116, opened 2018-08-24
v0.62.0 — Component fixes, updates and additions.

DamianMullins:
### Added
- Body object styles.
- `c-card` modifier.
- Optional `c-pageBanner` component.
- Rays images.

### Changed
- `c-mediaElement` image border radius.

### Fixed
- Extra space in utilities styles.
- Documentation links in components.

### Example of the rays in action

ezgif com-optimize 1


src/img/decoration/rays.svg
```diff
@@ -0,0 +1 @@
+<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 1272 72"><g fill="none" fill-rule="nonzero"><path fill="#FFF" d="M0 0l1272 72H0z" opacity=".3"/><path fill="#FFF" d="M0 24l1272 48H0z" opacity=".6"/><path fill="#F5F5F5" d="M0 48l1272 24H0z"/></g></svg>
```
ashleynolan: Are `rays` and `rays--small` just white diagonal images?  Hard to see how it's used just from the rich diff (or what they look like).  An image of their usage may be useful just to see them in action.

src/scss/components/_media-element.scss
```diff
@@ -10,6 +10,7 @@
 $mediaElement-img-width         : 50px;
 $mediaElement-img-width--mid    : 78px;
 $mediaElement-img-width--wide   : 67px;
+$mediaElement-img-borderRadius  : spacing() !default;
```
ashleynolan: Do you definitely want this borderRadius hooked into the spacing?  Might be better to change it so they aren't linked as if the spacing changes, I'm guessing you wouldn't want/expect this to update?

src/scss/components/_page-banner.scss
```diff
@@ -0,0 +1,60 @@
...
+            width: 100%;
+
+            @include media('>mid') {
+                background-image: inline('#{$img-path}/decoration/rays.svg');
+            }
+        }
+    }
+}
+
+    .c-pageBanner--negativeBottom--aboveMid {
```
oanuna: indentation ?

src/scss/objects/_body.scss
```diff
@@ -0,0 +1,16 @@
+/**
+ * Body styles
+ * ===================================
+ */
+
+.o-body--bgWhite {
+    background-color: $white;
+}
+
+    .o-body--bgWhite--belowMid {
```
oanuna: indentation cc?

src/scss/components/_media-element.scss
DamianMullins: Makes sense, I saw `8px` and instantly jumped to the `spacing` helper!

src/img/decoration/rays.svg
DamianMullins: They are just diagonal images, yes. I'll attach an image to the PR description and follow this PR up with a new section in the Fozzie docs to cover them.

src/scss/components/_page-banner.scss
```diff
@@ -0,0 +1,63 @@
...
+                width: 100%;
+
+                @include media('>mid') {
+                    background-image: inline('#{$img-path}/decoration/rays--white.svg');
+                }
+            }
+        }
+    }
+
+        .c-pageBanner--negativeBottom--aboveMid {
```
ashleynolan: Would be good to add a comment saying what this is for.  Seems like it's moving moving something below it up, but use-case would be good to include

src/img/decoration/rays.svg
ashleynolan: 👍 Should they be part of `f-icons` potentially?  Just wondering as we've moved images out of fozzie in the past to keep them all in one place.

ashleynolan:
Just as an aside – should all icons live inside `f-icons`?  Wondering if we should be trying to put all icons in that repo, and have peerdependencies on that, or if there's value in keeping some icons in individual fozzie modules?

DamianMullins:
@ashleynolan I did consider putting the images into f-icons but they aren't technically icons so I wasn't sure. I'm happy to move them across though?

PR #120, opened 2018-08-29
v0.66.0 - update styles to target SVG inside of wrapped instead of cl…

sonnyprince:
[no visual change]
- update styles to target SVG inside of wrapped instead of class to work with inline-svg helper

## UI Review Checks


ashleynolan:
Would be good to comment these to say that a TODO should be to change this once the `inlineSVG` helper takes attributes.

sonnyprince:
@ashleynolan good idea - will do that now

ashleynolan:
Is there a linked documentation PR for these changes, or do they not need any updates?

PR #121, opened 2018-08-30
v0.67.0 - updates to listing item

sonnyprince:
_Update to listing tile._

Before:
screen shot 2018-08-30 at 10 26 27

After:
screen shot 2018-08-30 at 10 25 39

## UI Review Checks

## Browsers Tested

- [x] Chrome
- [x] Edge
- [x] Internet Explorer 11
- [x] Mobile

### DIFF
--- a/src/scss/components/_listings.scss
+++ b/src/scss/components/_listings.scss
@@ -13,7 +13,6 @@ $listing-borderColor            : $color-border;
 $listing-iconFill               : $grey--mid;
 $listing-labelBackground        : $grey--offWhite;
 $listing-labelBorder            : 2px;
-$listing-offlineBorderColor     : $grey--lighter;
 $listing-offlineBackground      : $grey--lightest;
 $listing-borderRadius           : 4px;
 $listing-img-borderRadius       : 2px;
@@ -22,7 +21,6 @@ $listing-promotionColor         : $orange;
 
 .c-listing {
     opacity: 1;
-    background: $listing-background;
     border-radius: $listing-borderRadius;
     transition: opacity 200ms ease, transform 200ms ease;
 
@@ -43,6 +41,7 @@ $listing-promotionColor         : $orange;
 }
 
 .c-listing--offline {
+    padding: spacing(x2);
     background: $listing-offlineBackground;
 }
 
@@ -51,6 +50,7 @@ $listing-promotionColor         : $orange;
         display: block;
         position: relative;
         padding: spacing(x2);
+        background: $listing-background;
         border-bottom: 1px solid $listing-borderColor;
 
         .c-listing &:last-child {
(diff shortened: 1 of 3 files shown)


src/scss/components/_listings.scss
```diff
@@ -6,20 +6,21 @@
...
+$listing-promoLabelBackground   : $purple--light;
+$listing-borderColor            : $color-border;
+$listing-iconFill               : $grey--mid;
+$listing-labelBackground        : $grey--offWhite;
+$listing-labelBorder            : 2px;
+$listing-offlineBackground      : $grey--lightest;
+$listing-borderRadius           : 4px;
+$listing-img-borderRadius       : 2px;
+$listing-textDividerBackground  : $grey--light;
+$listing-promotionColor         : $orange;
```
ashleynolan: Should probably be called `listing-promoIconColor` as otherwise this variable is a little vague

src/scss/components/_listings.scss
```diff
@@ -90,179 +103,168 @@ $listing-borderRadius       : 4px;
...
             }
 
-            @include media('>wide') {
-                right: 0;
-                width: 260px;
-                margin-top: 10px;
-                position: absolute;
-                bottom: spacing(x2);
-                padding-left: spacing(x2);
+            .c-listing-item-promotionText {
```
ashleynolan: Should be `.c-listing-item-promotion-text` (and doesn't need to be nested)

src/scss/components/_listings.scss
```diff
@@ -90,179 +103,168 @@ $listing-borderRadius       : 4px;
...
+            position: relative;
+            margin-left: spacing() / 2;
+            @include font-size(small, true);
         }
 
         .c-listing-item-text {
             margin-top: 0;
         }
 
+        .c-listing-item-textInline {
```
ashleynolan: This looks like it may be a modifier.  How is this used in markup?

src/scss/components/_listings.scss
```diff
@@ -90,179 +103,168 @@ $listing-borderRadius       : 4px;
...
         }
 
-        .c-listing-item-details {
-            width: 100%;
-            padding-left: 56px;
+        .c-listing-item-detailsRow-text {
+            margin-left: spacing();
+        }
+
+        .c-listing-item-promotion {
```
ashleynolan: What's the difference between this class and the `c-listing-item-promo`?

sonnyprince: the promotion is the tiles 20% off text. The promo is the alternative tile label "BTA winner 2018" but was labled promotion. 

I was blind! 

is `.c-listing-item-labelAlt` okay?

src/scss/components/_listings.scss
sonnyprince: fixed 👍

src/scss/components/_listings.scss
sonnyprince: removed - used the layout class `l-inline` instead 👍

src/scss/components/_listings.scss
sonnyprince: fixed 👍

src/scss/components/_listings.scss
ashleynolan: Makes sense – so they're both promotions of different things. I guess the BTA one could be an `award` and be labelled like that?  The 20% should likely be labelled `promo`

sonnyprince: yeah - that sounds good. Will update now.

sonnyprince: fixed 😃 👍

PR #126, opened 2018-08-31
v0.69.0 — Badge component updates.

DamianMullins:
### Added
- Added `c-badge--success`.
- Added `c-badge--important`.
- Added `c-badge--award`.
- Added `c-badge--angled`.

### Changed
- Updated `c-badge` padding.

### DIFF
--- a/src/scss/components/_badges.scss
+++ b/src/scss/components/_badges.scss
@@ -84,28 +84,26 @@ $badge-rounded-radius       : 14px !default;
 
         .c-badge--angled {
             display: inline-block;
-            overflow: hidden;
-            padding-left: 15px;
-            padding-right: 15px;
+            margin: 0 5px;
 
             &:before,
             &:after {
-                background-color: $white;
+                background-color: inherit;
                 content: '';
                 height: 100%;
                 position: absolute;
                 top: 0;
-                transform-origin: 100% 0px;
+                transform-origin: 100% 0;
                 transform: skew(-15deg);
-                width: 12px;
+                width: 8px;
             }
 
             &:before {
-                left: -5px;
+                left: 0;
             }
 
             &:after {
-                right: -12px;
+                right: -8px;
             }
         }
 
--- a/src/scss/components/_badges.scss
+++ b/src/scss/components/_badges.scss
@@ -88,23 +88,24 @@ $badge-rounded-radius       : 14px !default;
             padding-left: 15px;
             padding-right: 15px;
 
-            &:before {
+            &:before,
+            &:after {
+                background-color: $white;
                 content: '';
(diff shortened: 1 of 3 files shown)


src/scss/components/_badges.scss
```diff
@@ -11,51 +11,101 @@
...
 
+        .c-badge--angled {
+            display: inline-block;
+            overflow: hidden;
+            padding-left: 15px;
+            padding-right: 15px;
+
+            &:before {
+                content: '';
+                border: 25px solid $white;
```
sonnyprince: is there anything we can do here for the border colour to inherit from the parent? 
Just thinking if the parents background is black: 
screen shot 2018-08-31 at 13 33 36

DamianMullins: Yeah, that would be good, I couldn't figure out a way to do that though. Any ideas?

sonnyprince: hmm, maybe instead of using border we could skew the `before` and `after`, like so:

```
&:after {
                content: '';
                width: 15px;
                height: 100%;
                background: inherit;
                position: absolute;
                right: -5px;
                top: 0;
                transform: skewX(-20deg);
            }
```

creates:
screen shot 2018-08-31 at 14 00 00

this _might_ work?

DamianMullins: OK, so this works but isn't quite ideal

```css
.c-badge--angled {
    display: inline-block;
    transform: skew(-15deg);

    span {
        display: block;
        transform: skew(15deg);
    }
}
```

```html
<p class="c-badge c-badge--large c-badge--angled c-badge--alert"><span>Alert</span></p>
```

Note the nested span.

sonnyprince: If you remove the internal padding and add a little margin to allow the skew, i get this:

screen shot 2018-08-31 at 14 39 43

from:
`<div class="c-badge c-badge--angled c-badge--award">hey</div>`

(ignore the colours, i fudged them for a test)

```
.c-badge--angled {
            display: inline-block;
            margin: 0 10px;

            &:before,
            &:after {
                background-color: inherit;
                content: '';
                height: 100%;
                position: absolute;
                top: 0;
                transform-origin: 100% 0px;
                transform: skew(-15deg);
                width: 12px;
            }
```

sonnyprince: that does add more padding, within the badge, than we probably want thought 🤔 - feel free to ignore my last comment 😄

DamianMullins: That's it! 🏁 

I owe you many 🍻's! 

screen shot 2018-08-31 at 14 52 04

PR #129, opened 2018-09-03
v0.73.0 — Added Optional `c-menu` component.

DamianMullins:
Queued behind #130.

### Added
- Optional `c-menu` component.

### Changed
- `l-inlined` extended so that it can be referenced as a placeholder.
- Applied `z-index` to `c-pageBanner` component.
- `$truncation-boundary` parameter is now optional on the `truncate` mixin.

fozzie-menu-component

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,19 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html)
 
 
+v0.73.0
+------------------------------
+*September 4, 2018*
+
+### Changed
+- Changed fullScreenPopOver JS to `is-open`
+- Optional `c-menu` component.
+
+### Changed
+- `l-inlined` extended so that it can be referenced as a placeholder.
+- `$truncation-boundary` parameter is now optional on the `truncate` mixin.
+
+
 v0.72.0
 ------------------------------
 *September 4, 2018*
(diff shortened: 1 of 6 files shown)


src/scss/components/_menu.scss
```diff
@@ -0,0 +1,69 @@
...
+
+            &.has-icon {
+                @extend %l-inlined;
+
+                .c-menu-link {
+                    padding-right: spacing() / 2;
+                }
+            }
+        }
+            .c-menu-item:active,
```
sonnyprince: could these use the `&` and be nested within `.c-menu-item` - similar to line 58-61?

DamianMullins: Yes, I'd also have to pull in the `.c-menu-item--active` class as well unless I want to duplicate styles? Placeholders won't work here to reduce duplication either as it adds the styles higher up the file which means the 2px border never has precedence.

sonnyprince: ahh - that makes sense then!
LGTM then 👍

sonnyprince:
@DamianMullins rebase and this is GTG 👍

CHANGELOG.md
```diff
@@ -4,6 +4,18 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html)
 
 
+v0.73.0
```
ashleynolan: Where's v0.72?  Not sure if the versions are out of sync a bit…

sonnyprince: 0.72.0 was released ~hour ago :)

ashleynolan: 🤦‍♂️

PR #131, opened 2018-09-04
v0.74.0 - listing badge style amends 

sonnyprince:
- Removed styles from listing component so we can utilise c-badge
- Added `c-listing-item-premier` (Local legend badge) styling

screen shot 2018-09-04 at 16 20 03

## Browsers Tested

- [x] Chrome
- [x] Edge
- [x] Internet Explorer 11
- [x] Mobile

### DIFF
--- a/src/scss/components/_badges.scss
+++ b/src/scss/components/_badges.scss
@@ -83,6 +83,10 @@ $badge-angled-rounded-radius: 2px !default;
             @include font-size(base--scaleUp);
         }
 
+        .c-badge--noPad {
+            padding: 0;
+        }
+
         .c-badge--angled {
             display: inline-block;
             margin: 0 5px;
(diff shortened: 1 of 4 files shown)


src/scss/components/_badges.scss
```diff
@@ -82,6 +83,10 @@ $badge-rounded-radius       : 14px !default;
             @include font-size(base--scaleUp);
         }
 
+        .c-badge--noPad {
+            padding: 0;
```
ashleynolan: What does the badge look like without padding?  Just tried it in the component library and it looks pretty strange without any padding at all (know this is a design question, but it's more to do with consistency across SERP & Menu usage)

sonnyprince: This is for removing the padding of the badges with the span divider:

screen shot 2018-09-05 at 09 59 47
screen shot 2018-09-05 at 09 59 51

ashleynolan: Ah, ok – could you make sure this modifier gets added to the badge page in the docs too?

sonnyprince: no problem!

sonnyprince: https://github.com/justeat/global-component-library/pull/48

PR #136, opened 2018-09-06
v0.78.0 — Card component section child styles.

DamianMullins:
### Added
- Card component section child styles.

card-section--collapsible

### DIFF
--- a/src/scss/components/_cards.scss
+++ b/src/scss/components/_cards.scss
@@ -9,14 +9,18 @@
  * https://fozzie.just-eat.com/styleguide/ui-components/cards
  */
 
-$card-bgColor--active           : $white !default;
-$card-bgColor--disabled         : $grey--lightest !default;
-$card-tooltip-width             : 10px !default;
-$card-arrow-bottom-position     : 0 !default;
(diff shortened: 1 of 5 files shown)


src/scss/components/_cards.scss
```diff
@@ -62,5 +66,68 @@ $card-borderColor               : $grey--lightest !default;
...
+
+            .c-card-section--collapsed {
+                height: 50px;
+                overflow: hidden;
+
+                &.c-card-section {
+                    margin-bottom: -#{spacing(x2)};
+                }
+
+                p {
```
sonnyprince: Should add a class and target that instead, something like `.c-card-section-text` - just incase the the `p` changes in the future :)

DamianMullins: The thing is, the text comes from the back-end somewhere so we can't easily add a class to the content.

sonnyprince: ahh - bummer. No worries then 😄

DamianMullins: I'll keep your comment in mind though as there may be an opportunity to fix that text from within the menu vertical.

PR #138, opened 2018-09-10
v0.80.0 — Card styles updated.

DamianMullins:
### Added
- Card no padding modifier.
- Card highlight modifier.
- Card bottom section modifier.
- Animation to card section chevron.
- 0.5 spacing value.

### Changed
- Updated modules to use new 0.5 spacing value.

### Fixed
- Card section margin only applied on larger screens.

screen shot 2018-09-10 at 13 00 35

### DIFF
--- a/src/scss/components/_cards.scss
+++ b/src/scss/components/_cards.scss
@@ -19,6 +19,8 @@ $card-radius--large                             : 8px !default;
 $card-borderColor                               : $grey--lightest !default;
 $card-section-margin                            : spacing(x4) !default;
 $card-section-margin--large                     : spacing(x8) !default;
+$card-section-highlight-backgroundColor         : $orange;
+$card-section-highlight-color                   : $white;
 $card-section-collapsible-paddingRight          : 42px;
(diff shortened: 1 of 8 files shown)


src/scss/components/_cards.scss
```diff
@@ -78,6 +84,21 @@ $card-section-collapsible-paddingRight--aboveMid: 60px;
...
         }
 
+        // used to create a section at the end of a card which has a background colour applied
+        .c-card-section--highlight {
+            background-color: $card-section-highlight-backgroundColor;
+            border-bottom-right-radius: $card-radius;
+            border-bottom-left-radius: $card-radius;
+            color: $card-section-highlight-color;
+            margin-top: spacing(x2);
+            padding: (spacing() / 2) spacing();
```
ashleynolan: Reckon it's worth adding a `0.5` spacing option?  Know this is being used in a few places now.

DamianMullins: Cool, done. Didn't know you specify keys like that 😄

PR #141, opened 2018-09-11
v0.83.0 - addition of text colour util classes

sonnyprince:
_addition of text colour util classes._

## Browsers Tested

- [x] Chrome
- [x] Edge
- [x] Mobile

### DIFF
--- a/src/scss/trumps/_utilities.scss
+++ b/src/scss/trumps/_utilities.scss
@@ -146,14 +146,6 @@
     color: $color-text--warning;
 }
 
-.u-color-headings {
-    color: $color-headings;
-}
-
-.u-color-headings--highlight {
-    color: $color-headings--highlight;
-}
-
 //
 // Spacing Utilities
 // ==========================================================================


package.json
```diff
@@ -2,7 +2,7 @@
   "name": "@justeat/fozzie",
   "title": "Fozzie – Just Eat UI Web Framework",
   "description": "UI Web Framework for the Just Eat Global Platform",
-  "version": "0.80.0",
+  "version": "0.83.0",
```
kevinrodrigues: Should this jump to 3 versions?

kevinrodrigues: Ah actually you have two other PRs before, ignore :)

sonnyprince: i have pr's ope for 81 and 82 - will be merged / rebased in that order 👍

src/scss/trumps/_utilities.scss
```diff
@@ -122,6 +122,38 @@
...
+
+.u-color-text--danger {
+    color: $color-text--danger;
+}
+
+.u-color-text--warning {
+    color: $color-text--warning;
+}
+
+.u-color-headings {
```
ashleynolan: Definitely understand needing the classes above – with the headings ones, shouldn't they already have this colour defined if they are a heading?

There are some utility classes that make other elements look like headings, so may be best to just use those, otherwise it's kinda encouraging people to make headings that don't follow the design conventions (and just have the colour)

sonnyprince: yeah that's true! I'll remove those two.

PR #144, opened 2018-09-11
v0.86.0 - addition of cursor pointer to o-btnLink on hover

sonnyprince:
- addition of cursor pointer to o-btnLink on hover, thats all.

PR #145, opened 2018-09-12
v0.87.0 — Updating card and utilities.

DamianMullins:
To be merged after #144 

### Added
- Card component normalise links modifier.
- Position sticky utility helper.
- Padding utility helpers.

### Changed
- Applied `!important` declaration to spacing utilities to ensure that they always trump other styles.
- Badge component text never wraps.
- Increased spacing between badges.
- Card components display as blocks.
- Card component comments updated.

### Fixed
- Overlap of angled badge section over text.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,14 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html)
 
 
+v0.86.0
+------------------------------
+*September 12 2018*
+
+### Added
+- Added `cursor: pointer` to the hover of `o-btnLink`.
+
+
 v0.85.0
 ------------------------------
 *September 11 2018*
--- a/package.json
+++ b/package.json
@@ -2,7 +2,7 @@
   "name": "@justeat/fozzie",
   "title": "Fozzie – Just Eat UI Web Framework",
   "description": "UI Web Framework for the Just Eat Global Platform",
-  "version": "0.85.0",
+  "version": "0.86.0",
   "main": "dist/js/index.js",
   "files": [
     "dist/js",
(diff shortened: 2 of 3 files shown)


src/scss/trumps/_utilities.scss
```diff
@@ -109,7 +114,7 @@
 }
 
 .u-textPad {
-    padding-left: spacing(x4);
+    padding-left: spacing(x4) !important;
```
sonnyprince: I wonder if we could include our utils last in the imports so we don't have to use `!important`'s 🤔
future talking point - not an issue for this PR.

oanuna: yeah is this !important really required?

oanuna: Actually should we really be using importants in any padding. if so then we're probably constructing our components wrong and it would be better to extend the class within the component i.e.

```scss
.my-class {
  @extend .u-textPad
}
```

DamianMullins: The issue is that when you have optional imports they _must_ be imported _after_ the main Fozzie import, which means that the component styles take precedence over the trumps.

I think it's one we could come back to in the future to discuss for sure 🚜

ashleynolan: @oanuna `!important` in this scenario is just enforcing the utility class really.  If people are overriding the styles provided by the utility class they shouldn't be using it 😄 

Agree used anywhere else it's a symptom of badly constructed components, but it kind of re-enforces how they should be applied here.

There's a good article on `!important` [not always being bad here](https://csswizardry.com/2016/05/the-importance-of-important/) (especially the section on utility classes).

DamianMullins:
Not sure what I've done to this branch but it says it was merged when it hasn't been? I'm going to create a fresh PR...

DamianMullins:
Recreated in #146.

PR #156, opened 2018-09-14
v0.95.0 - listing badge responsiveness

sonnyprince:
- addition of c-listing-item-badge to handle badge responsiveness

Before (on breakpoints between mid and wide):
screen shot 2018-09-14 at 21 21 32

After:
screen shot 2018-09-14 at 21 23 25

---

## Browsers Tested

- [x] Chrome
- [x] Edge
- [x] Internet Explorer 11
- [x] Mobile


src/scss/components/_listings.scss
```diff
@@ -164,6 +164,56 @@ $listing-promoIconColor         : $orange;
...
+            between mid and wide you will see:
+
+            item
+            item
+
+            The mixins are used to stop code duplication as each chunk of 
+            code needs to be within its own media mixin.
+            */
+
+            @mixin inlineSpan() {
```
ashleynolan: Only thing that would be better here is if these were named less tied to element names (as the span's may one day change in the future and then the name of the mixins would need updating too).

Looks good otherwise.

sonnyprince:
@80sgreek @ashleynolan rebase dismissed your ticks 😄 🙏

JonnyWideFoot:
Being as "@80sgreek @ashleynolan rebase dismissed your ticks" -- no significant chaneg since

PR #158, opened 2018-09-18
v0.97.0 - colour util

sonnyprince:
_Addition of colour utility class._

Addition of `.u-color-secondary`. It is the same color as `.u-color-link` but i thought it would be better for items that aren't a link but require that colour.


kevinrodrigues:
It's probably better to open these type of prs after the previous one merges.

sonnyprince:
@kevinrodrigues will do in the future 👍

src/scss/trumps/_utilities.scss
```diff
@@ -139,7 +139,8 @@
     color: $color-text;
 }
 
-.u-color-link {
+.u-color-link,
+.u-color-secondary {
```
markedup: Without a pre-existing `.u-color-primary` class, I'm not sure I understand what this is for, or how we should apply it in future; is there a better name for this?

sonnyprince: `-secondary` refers to directly to a colour variable within fozzie - as does `-link` ($color-link-default and $color-secondary (both use $blue)).

If the classes were not directly linked to a fozzie colour var I would agree :)

markedup: Okay, that connection isn't obvious from this instance, so I was concerned it wouldn't be obvious for other developers. I'll approve in this instance though.

sonnyprince:
@ashleynolan It would yes, but i was just trying to make non-link html not have CSS that refers to links

PR #161, opened 2018-09-19
v0.98.0 - update hover states on pop over action button

sonnyprince:
- update hover states on pop over action button

Before:
screen shot 2018-09-19 at 10 40 20

After:
screen shot 2018-09-19 at 10 40 04

## UI Review Checks

## Browsers Tested

- [x] Chrome
- [x] Edge
- [x] Internet Explorer 11
- [x] Mobile

### DIFF
--- a/src/scss/components/_fullscreen-pop-over.scss
+++ b/src/scss/components/_fullscreen-pop-over.scss
@@ -58,7 +58,13 @@ $fullScreenPopOver-shadow-color      : rgba($black, 0.12);
     .c-fullScreenPopOver-header-actionSecond {
         top: 50%;
         position: absolute;
+        text-decoration: none;
         transform: translateY(-50%);
+
+        &:hover,
+        &:focus {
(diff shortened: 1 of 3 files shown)


src/scss/components/_fullscreen-pop-over.scss
```diff
@@ -58,7 +58,13 @@ $fullScreenPopOver-shadow-color      : rgba($black, 0.12);
     .c-fullScreenPopOver-header-actionSecond {
         top: 50%;
         position: absolute;
+        text-decoration: none;
```
kevinrodrigues: Hmm from an accessibility side of things is this ok? The underline visually makes it look like links IMO.

sonnyprince: Agree - this is something that Ash and Pieter-Jan Voskamp have covered before. 
I think Design / UX are going to do a a full review in the coming weeks but for now all is okay.

PR #163, opened 2018-09-24
v0.100.0 - addition of stopFOIT module and unit tests

sonnyprince:
### Added
- Stop FOIT JS module to reduce meaningful paint with webfonts

**Before (slow 3g):**
without-preload

**After (slow 3g):**
with-preload-font

### Changed
- formToggle count text size

  - [x] HTML is valid

## Browsers Tested

- [x] Chrome
- [x] Edge
- [x] Internet Explorer 11
- [x] Mobile

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 v0.100.0
 ------------------------------
-*September 24, 2018*
+*September 26, 2018*
 
 ### Added
 - Stop FOIT JS module to reduce meaningful paint with webfonts
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,17 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html)
 
 
+v0.100.0
+------------------------------
+*September 24, 2018*
+
+### Added
+- Stop FOIT JS module to reduce meaningful paint with webfonts
+
+### Changed
+- formToggle count text size
+
+
 v0.99.2
 ------------------------------
 *September 25, 2018*
--- a/src/js/modules/stopFoit/test/index.test.js
+++ b/src/js/modules/stopFoit/test/index.test.js
@@ -2,7 +2,17 @@ import FontFaceObserver from 'fontfaceobserver';
 import TestUtils from 'js-test-buddy';
 import { stopFoit } from '../index';
 
+jest.mock('fontfaceobserver');
+
 describe('module stopFoit', () => {
+    beforeAll(() => {
+        FontFaceObserver.mockImplementation(() => ({
+            load: () => new Promise(resolve => {
+                resolve();
+            })
+        }));
+    });
+
     it('adds `has-fontsLoaded--headings` to the body when Ubunto webfont has loaded', () => {
         // Arrange
         TestUtils.setBodyHtml(`
@@ -13,13 +23,13 @@ describe('module stopFoit', () => {
         stopFoit();
 
         // Assert
-        const headingFont = new FontFaceObserver('Ubuntu');
-        headingFont.load().then(() => {
+        setTimeout(() => {
+            expect(FontFaceObserver).toBeCalled();
             expect(document.body.classList.contains('has-fontsLoaded--headings')).toBe(true);
-        });
+        }, 100);
     });
 
-    it('adds `has-fontsLoaded--base` to the body when Ubunto webfont has loaded', () => {
+    it('adds `has-fontsLoaded--base` to the body when Hind Vadodara webfont has loaded', async () => {
         // Arrange
         TestUtils.setBodyHtml(`
             <link href="https://fonts.googleapis.com/css?family=Hind+Vadodara" rel="stylesheet">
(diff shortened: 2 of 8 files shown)


src/js/modules/stopFoit/test/index.test.js
```diff
@@ -0,0 +1,37 @@
+import FontFaceObserver from 'fontfaceobserver';
+import TestUtils from 'js-test-buddy';
+import { stopFoit } from '../index';
+
+describe('module stopFoit', () => {
```
kevinrodrigues: Missing the tests for the `catch` scenario. :)

CHANGELOG.md
```diff
@@ -4,6 +4,16 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html)
 
 
+v0.100.0
+------------------------------
+*September 24, 2018*
```
ashleynolan: Bump

src/scss/settings/_variables.scss
```diff
@@ -61,6 +61,12 @@ $font-family-base           : 'Hind Vadodara', 'Helvetica Neue', Helvetica, sans
 $font-family-headings       : 'Ubuntu'; // we need to define some fallbacks here!
 $font-family-mono           : Menlo, Monaco, 'Courier New', monospace;
 
+// Fonts for preload
+// ==========================================================================
+// base fonts and headings have a preload system font, when the webfonts are loaded we style with JustEat fonts.
+
+$font-family-preload: Helvetica, sans-serif;
```
ashleynolan: Could add 'Helvetica Neue' here as well potentially, as that's a base system font for MacOS (and will just fallback automatically to Helvetica if not present on Windows.

src/scss/base/_typography.scss
```diff
@@ -24,24 +24,32 @@ html {
...
+    font-family: $font-family-preload;
     @include font-size(base);
     color: $color-text;
     font-weight: $font-weight-base;
 
     text-rendering: optimizelegibility;
     -webkit-font-smoothing: antialiased;
     -moz-font-smoothing: antialiased;
+
+    .has-fontsLoaded--base & {
```
ashleynolan: I think I'd rather this work the other way around to how it is atm, as it'll be easier to maintain going forward (as we'd have to add this modifier everywhere a font-family is defined doing this).

Instead, we could apply the `$font-family-preload;` to a set of selectors:

```
.is-fontsLoading {
  body,
  p,
  h1… {
    font-family: $font-family-preload !important;
  }
}
```

and remove the `is-fontsLoading` class once they're loaded which will return the styles to the default as defined.

src/js/modules/stopFoit/test/index.test.js
```diff
@@ -0,0 +1,86 @@
...
+
+        it('causes FontFaceObserver to reject', () => {
+            // Arrange
+            TestUtils.setBodyHtml('');
+
+            // Act
+            stopFoit();
+
+            // Assert
+            expect(FontFaceObserver.rejects);
```
kevinrodrigues: This expect should have a matcher function:

```
expect(FontFaceObserver.rejects).toHaveBeenCalled()
```

src/scss/base/_typography.scss
```diff
@@ -18,6 +18,36 @@
...
+
+.is-fontsLoading--base {
+
+    body,
+    p {
+        font-family: $font-family-preload !important;
+    }
+}
+
+.is-fontsLoading--heading {
```
DamianMullins: Can you not add these styles into the existing section below using something like `.is-fontsLoading--heading &`? Otherwise, the list of selectors will have to be maintained in two places.

sonnyprince: 🤦‍♂️

DamianMullins:
@sonnyprince Is there potentially something we could do with the [`font-display` property](https://developers.google.com/web/updates/2016/02/font-display) as well to help with this?

sonnyprince:
@DamianMullins - i did look into `font-display` but google doesnt support this yet, it needs to be within the font-face that they serve. 
Apparently the google lighthouse team are on to the webfont team about this (i read it somewhere on github before even looking into using fontface observer)

src/scss/base/_typography.scss
```diff
@@ -60,13 +72,18 @@ h6,
 .epsilon {
     color: $color-headings;
     margin: 0;
-    font-family: $font-family-headings;
-    font-weight: $font-weight-headings;
     margin-bottom: 0;
+    font-family: $font-family-headings;
+    font-weight: $font-weight-base;
```
DamianMullins: I think this needs to be `$font-weight-headings`. It's messing up the heading weights for me.

src/scss/base/_typography.scss
```diff
@@ -60,13 +72,18 @@ h6,
...
 
     small {
         font-weight: normal;
     }
+
+    .is-fontsLoading--heading & {
+        font-family: $font-family-preload !important;
+        font-weight: $font-weight-bold;
```
DamianMullins: Don't think this one is needed? Should pick up the weight from the styles above?

src/scss/base/_typography.scss
```diff
@@ -32,6 +36,10 @@ body {
     text-rendering: optimizelegibility;
     -webkit-font-smoothing: antialiased;
     -moz-font-smoothing: antialiased;
+
+    .is-fontsLoading--base & {
```
ashleynolan: More of an optimisation thing – better to pull this out so that all entries use the same styles:

```
html,
p {
  .is-fontsLoading--base & {
      font-family: $font-family-preload !important;
  }
}
```

PR #171, opened 2018-10-05
v1.3.0 - within breakpoint method

sonnyprince:
Extension of current breakpointHelper JS to include `withinBreakpoint`.

`withinBreakpoint` is build to mimic our current SCSS breakpoint media queries. 

`withinBreakpoint('>mid')` and will return true or false.

---

## UI Review Checks

## Browsers Tested

- [x] Chrome
- [x] Edge
- [x] Internet Explorer 11
- [x] Mobile

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,8 +8,8 @@ v1.3.0
 ------------------------------
 *October 5, 2018*
 
-- Added `withinBreakpoint` method to breakpoint helper for breakpoint queries that match our SCSS style
-- Added associated unit tests for `withinBreakpoint` method 
+- Added `isWithinBreakpoint` method to breakpoint helper for breakpoint queries that match our SCSS style
+- Added associated unit tests for `isWithinBreakpoint` method 
 
 
 v1.2.0
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,14 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html)
 
 
+v1.3.0
+------------------------------
+*October 5, 2018*
+
+- Added `withinBreakpoint` method to breakpoint helper for breakpoint queries that match our SCSS style
+- Added associated unit tests for `withinBreakpoint` method 
+
+
 v1.2.0
 ------------------------------
 *October 3, 2018*
@@ -11,6 +19,7 @@ v1.2.0
 ### Fixed
 - Fixed spacing on `c-listing` component
 - Fixed display on `c-overflowCarousel` component
+
 
 v1.1.0
 ------------------------------
--- a/src/js/modules/breakpointHelper.js
+++ b/src/js/modules/breakpointHelper.js
@@ -67,7 +67,7 @@ export const getCurrentScreenWidth = () => {
     return false;
 };
 
-export const withinBreakpoint = breakpointString => {
+export const isWithinBreakpoint = breakpointString => {
     const operatorRegex = /[<>=]+/;
     const operatorMatch = breakpointString.match(operatorRegex);
     const operator = operatorMatch ? operatorMatch[0] : '';
--- a/src/js/modules/breakpointHelper.js
+++ b/src/js/modules/breakpointHelper.js
@@ -4,8 +4,6 @@
  * @module breakpointHelper
  */
 
-let bps;
-
 export const getBreakpoints = () => {
     const output = {};
 
@@ -36,17 +34,23 @@ export const getBreakpoints = () => {
     }, output); // <- initial value
 };
 
+export const createBreakpointArray = breakpoints => {
+    // Order the breakpoints from widest to narrowest,
+    // takes the form [['narrow', '414px'], [...etc]]
+    const bps = [];
+    Object.keys(breakpoints).forEach(key => {
+        bps.unshift([key, breakpoints[key]]);
+    });
+
+    return bps;
+};
+
 export const getCurrentScreenWidth = () => {
     const currentWidth = window.innerWidth;
 
     const breakpoints = getBreakpoints();
 
-    // Order the breakpoints from widest to narrowest,
-    // takes the form [['narrow', '414px'], [...etc]]
-    bps = [];
-    Object.keys(breakpoints).forEach(key => {
-        bps.unshift([key, breakpoints[key]]);
-    });
+    const bps = createBreakpointArray(breakpoints);
 
     for (let i = 0; i < bps.length; i++) {
         // Loops through the breakpoints (in descending order)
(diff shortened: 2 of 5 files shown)


src/js/modules/breakpointHelper.js
```diff
@@ -60,3 +62,57 @@ export const getCurrentScreenWidth = () => {
...
+    const currentScreenWidth = window.innerWidth;
+    let match;
+
+    // We fire getCurrentScreenWidth to set the bp array we loop through to get the PX values
+    getCurrentScreenWidth();
+
+    // We loop through the breakpoint array until we get a match.
+    // If we match we return the px value as an int. If we do not match we return false
+    const breakpointToPX = bp => {
+        for (let i = 0; i < bps.length; i++) {
```
kevinrodrigues: `ForEach`?

src/js/modules/breakpointHelper.js
```diff
@@ -60,3 +62,57 @@ export const getCurrentScreenWidth = () => {
     // If no breakpoints have been set
     return false;
 };
+
+export const withinBreakpoint = breakpointString => {
+    const operatorMatch = breakpointString.match(/[<>=]+/);
```
DamianMullins: Can you move the regex out into a variable so that it isn't repeated?

src/js/modules/breakpointHelper.js
```diff
@@ -60,3 +62,57 @@ export const getCurrentScreenWidth = () => {
...
+
+    // If the breakpoint passed in does not match any we return false;
+    if (breakpointInPX === false) {
+        return false;
+    }
+
+    // We match our passed in operator and execute a sum: current screen width [Passed operator] [Passed breakpoint in PX]
+    switch (operator) {
+        case '>':
+            match = currentScreenWidth > breakpointInPX;
```
DamianMullins: Can you just return the result here, rather than assign it to `match` and return later? That way you can discard the `match` variable altogether.

src/js/modules/breakpointHelper.js
```diff
@@ -60,3 +62,57 @@ export const getCurrentScreenWidth = () => {
...
+            }
+        }
+        return false;
+    };
+
+    const breakpointInPX = breakpointToPX(breakpoint);
+
+    // If the breakpoint passed in does not match any we return false;
+    if (breakpointInPX === false) {
+        return false;
```
kevinrodrigues: Do you need to return false here? `breakpointToPX` already returns false you could just `return` instead?

src/js/modules/breakpointHelper.js
```diff
@@ -60,3 +62,57 @@ export const getCurrentScreenWidth = () => {
...
+                return parseInt(bps[i][1], 10);
+            }
+        }
+        return false;
+    };
+
+    const breakpointInPX = breakpointToPX(breakpoint);
+
+    // If the breakpoint passed in does not match any we return false;
+    if (breakpointInPX === false) {
```
kevinrodrigues: `if (!breakpointInPX)`

src/js/test/index.test.js
```diff
@@ -86,3 +86,144 @@ describe('currentScreenWidth', () => {
         expect(width).toBe(false);
     });
 });
+
+describe('withinBreakpoint', () => {
+    it('should return true if I pass in `>narrow` when my screen width is larger than 414', () => {
+        // Arrange
+        TestUtils.setBodyHtml(`
```
DamianMullins: Extract the `setBodyHtml` function call out into a `beforeEach` as it is repeated in every test below?

src/js/test/index.test.js
```diff
@@ -86,3 +86,144 @@ describe('currentScreenWidth', () => {
...
+        window.innerWidth = 414;
+
+        // Act
+        const breakpointMatch = withinBreakpoint('=narrow');
+
+        // Assert
+        expect(breakpointMatch).toBe(true);
+    });
+
+    it('should return false if the passed operator is not accepted', () => {
```
DamianMullins: Is it worth testing for a value without a prefix as well? e.g. `withinBreakpoint('narrow');`

src/js/modules/breakpointHelper.js
```diff
@@ -60,3 +62,49 @@ export const getCurrentScreenWidth = () => {
     // If no breakpoints have been set
     return false;
 };
+
+export const withinBreakpoint = breakpointString => {
```
ashleynolan: Think maybe `isWithinBreakpoint` would be a better name if it returns true/false.

Also, can you add a comment to the function, with what params it takes as input?

src/js/test/index.test.js
```diff
@@ -86,3 +86,144 @@ describe('currentScreenWidth', () => {
         expect(width).toBe(false);
     });
 });
+
+describe('withinBreakpoint', () => {
```
kevinrodrigues: It's always good to check that this method exists first before testing the implementation.

```
it('should exist', () => {
  expect(withinBreakpoint).toBeDefined()
})
```

src/js/modules/breakpointHelper.js
```diff
@@ -41,7 +43,7 @@ export const getCurrentScreenWidth = () => {
 
     // Order the breakpoints from widest to narrowest,
     // takes the form [['narrow', '414px'], [...etc]]
-    const bps = [];
+    bps = [];
```
ashleynolan: Rather than pulling this out to be globally available, would be tidier to pull this out into a function that gets called and returns a set of breakpoints that can be called here and in `withinBreakpoint` when needed.

src/js/modules/breakpointHelper.js
```diff
@@ -60,3 +62,49 @@ export const getCurrentScreenWidth = () => {
...
+
+export const withinBreakpoint = breakpointString => {
+    const operatorRegex = /[<>=]+/;
+    const operatorMatch = breakpointString.match(operatorRegex);
+    const operator = operatorMatch ? operatorMatch[0] : '';
+    const [, breakpoint] = breakpointString.split(operatorRegex);
+    const currentScreenWidth = window.innerWidth;
+
+    // We fire getCurrentScreenWidth to set the bp array we loop through to get the PX values
+    getCurrentScreenWidth();
```
ashleynolan: Simialr to comment above – should pull out this functionality so you don't have to call the whole `getCurrentScreenWidth` function just to get the breakpoints array required here.

src/js/test/index.test.js
```diff
@@ -86,3 +86,117 @@ describe('currentScreenWidth', () => {
...
```
kevinrodrigues: `withinBreakpoints` invokes `getCurrentScreenWidth` to set the `bps` it's worth writing a test to make sure this gets called.

src/js/modules/breakpointHelper.js
sonnyprince: I get shouted at by the linter for inconsistent returns if i dont

CHANGELOG.md
```diff
@@ -4,6 +4,14 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html)
 
 
+v1.3.0
+------------------------------
+*October 5, 2018*
+
+- Added `withinBreakpoint` method to breakpoint helper for breakpoint queries that match our SCSS style
```
ashleynolan: Need to update to `isWithinBreakpoint` here and on line below

sonnyprince: done! 👍

PR #172, opened 2018-10-05
v1.4.0 — Added toast component and updated styles.

DamianMullins:
To go out after #171 

### Added
- Optional `c-toast` component.

### Changed
- `c-menu` active styles applied to link rather than list-item.
- `c-pagebanner` styles updated.
- `zIndex` map value for `low` changed.
- Sticky utility classes updated.

oct-05-2018 11-47-54


src/scss/components/_toast.scss
```diff
@@ -0,0 +1,91 @@
...
+$toast-bgColor--alert   : $red--dark;
+
+@mixin toast() {
+
+        @keyframes toastSlideUp {
+            0% {
+                top: 0;
+            }
+            100% {
+                top: -90px;
```
DamianMullins: The heights are currently fixed so that the CSS animation transitions can work. I'm open to suggestions on ways to do this better.

sonnyprince:
Has documentation been created yet? the link (https://fozzie.just-eat.com/styleguide/ui-components/toast) is 404'ing for me.

If not could we get a screenshot?

src/scss/components/_toast.scss
```diff
@@ -0,0 +1,91 @@
+/**
+ * Toast Component
+ * =================================
+ * Used for ...
```
sonnyprince: Could you update the `...`

src/scss/components/_toast.scss
sonnyprince: would a `translateY` work better than fixed pixels here? 🤔

DamianMullins:
@sonnyprince The docs page hasn't been published yet, I'm working on it at the moment. Also, I've added an animation of the element in action.

src/scss/components/_toast.scss
DamianMullins: Done.

src/scss/components/_toast.scss
```diff
@@ -0,0 +1,91 @@
...
+ * =================================
+ * Used to display text in a pop-up style notification.
+ *
+ * *The `c-toast` component is an optional mixin within Fozzie — if you'd like to use it in your project you can include it by adding `@include toast();` into your SCSS dependencies file.*
+ *
+ * Documentation:
+ * https://fozzie.just-eat.com/styleguide/ui-components/toast
+ */
+
+$toast-radius           : 8px !default;
```
ashleynolan: Is the `!default` used anywhere atm?

src/scss/components/_toast.scss
ashleynolan: Would be good to pull this out into a variable too, as would save having to update the value in multiple places if it does change.

You could link it to the height of the `c-toast` element then too if you know it always needs to pop up `5px` more than the height you could use `$toast-height: 85px` and `+ 5` where `90px` is used atm

src/scss/settings/_variables.scss
```diff
@@ -125,7 +125,7 @@ $border-radius: 4px;
 
 $zIndex: (
     lowest  : -1,
-    low     : 10,
+    low     : 1,
```
ashleynolan: Does anything else use `low` atm?

If so, might be worth adding a new key value for `1`.  If this isn't an issue though ignore me though :)

src/scss/components/_toast.scss
DamianMullins: @sonnyprince `translateY` does work better as it has removed the jank from the animation, nice one!

@ashleynolan That def make sense, will update now.

DamianMullins: For future ref: https://www.paulirish.com/2012/why-moving-elements-with-translate-is-better-than-posabs-topleft/

src/scss/components/_toast.scss
DamianMullins: Was left in after rebasing v1 🤦‍♂️  Will update.

src/scss/settings/_variables.scss
DamianMullins: Nothing within Fozzie uses it right now. Are these values exposed out?

ashleynolan: Yeh – they can be used by other modules/projects.  I highly doubt it is though (or that changing it to `1` will have any impact).

If it's ok within Fozzie happy for this to go in as is.

PR #174, opened 2018-10-09
v1.6.0 - Added `overflow-y: hidden` to overflow carousel to stop X an…

sonnyprince:
- Added `overflow-y: hidden` to overflow carousel to stop X and Y scrolling.

## UI Review Checks

Before:
scroll-fix-before

After:
scroll-fiz-after

## Browsers Tested

- [x] Chrome
- [x] Edge
- [x] Internet Explorer 11
- [x] Mobile

### DIFF
--- a/src/scss/components/optional/_overflow-carousel.scss
+++ b/src/scss/components/optional/_overflow-carousel.scss
@@ -22,7 +22,8 @@
 
         .c-overflowCarousel-inner {
             @include media('<mid') {
-                overflow: scroll;
+                overflow-x: scroll;
+                overflow-y: hidden;
                 -webkit-overflow-scrolling: touch;
                 width: 100%;
                 &.is-active {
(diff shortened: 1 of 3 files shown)


src/scss/components/optional/_overflow-carousel.scss
```diff
@@ -22,7 +22,8 @@
 
         .c-overflowCarousel-inner {
             @include media('<mid') {
-                overflow: scroll;
+                overflow-x: scroll;
+                overflow-y: hidden;
```
ashleynolan: Out of interested – if you didn't specify `overflow-y: hidden;`, would this still work? (as it wouldn't be `scroll` by default if not set, it'd be `visible`)

sonnyprince: just tested and removing `overflow-y: hidden;` re-creates the issue 😢

ashleynolan: Weird. CSS… :smil

PR #177, opened 2018-10-30
v1.9.0 - Star rating size issue on high DPR devices

sonnyprince:
- Move small star rating styles into mixin to reduce duplication
- Addition of pixel density ratio media query and override of spacing var within

On high (3+) pixel density ratio devices the small sized ratings would render ~ 6.1 stars

Before:
screen shot 2018-10-30 at 15 42 57

After:
screen shot 2018-10-30 at 15 48 12

## Browsers Tested

- [x] Mobile

### DIFF
--- a/src/scss/components/optional/_ratings.scss
+++ b/src/scss/components/optional/_ratings.scss
@@ -17,6 +17,11 @@ $star-size--medium      : 28px;
 $star-size--large       : 42px;
 $star-spacing           : $star-count;
 
+@mixin smallStarSizing() {
+    height: $star-size--small;
+    width: ($star-size--small * $star-count) + $star-spacing;
+}
+
 @mixin rating() {
 
     .c-rating {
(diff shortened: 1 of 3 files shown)


src/scss/components/optional/_ratings.scss
```diff
@@ -29,8 +34,14 @@ $star-spacing           : $star-count;
     }
 
     .c-rating--small {
-        height: $star-size--small;
-        width: ($star-size--small * $star-count) + $star-spacing;
+
+        @include smallStarSizing();
+
+        @media only screen and (-webkit-device-pixel-ratio : 3) {
```
ashleynolan: You can access this through `@include-media` – [see the docs here](https://include-media.com/documentation/#variable-media-expressions)

ashleynolan: Think it would be something like:

```
@include media('retina3x') { } 
```

PR #199, opened 2019-01-29
v1.25.0 — Updated component styles.

DamianMullins:
### Changed
- Updated card component styles.
- Updated page banner component styles.

### DIFF
--- a/src/scss/components/_cards.scss
+++ b/src/scss/components/_cards.scss
@@ -62,6 +62,20 @@ $card-section-collapsed-height--aboveMid        : 50px;
         padding-right: spacing(x2);
     }
 
+        .c-card--negativeSpacing--belowMid {
+            @include media('<mid') {
+                margin-left: -#{spacing(x2)};
+                margin-right: -#{spacing(x2)};
+            }
+        }
+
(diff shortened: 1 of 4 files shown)


src/scss/components/_cards.scss
```diff
@@ -62,6 +62,20 @@ $card-section-collapsed-height--aboveMid        : 50px;
         padding-right: spacing(x2);
     }
 
+        .c-card--negativeSpacing--belowMid {
+            @include media('<mid') {
+                margin-left: -#{spacing(x2)};
```
simonsandersje: Do you need the `#{...}` ?

DamianMullins: Yes, otherwise the compiled CSS comes out as `margin-left: -spacing();`. It needs the interpolation in order to compile the correct value with the minus sign 🐱

kevinrodrigues: If you put a space between the minus and function it works. But not a biggy here, just good to know, the interpolation is mostly best used for multiplication and adding variable and numbers together. :)

DamianMullins: Adding parenthesis around it works as well (`margin-left: (-spacing(x2));`).

PR #200, opened 2019-01-31
Wh 518 cookie banner

jinwood:
Add styles for cookie link, and shrink text size.


src/scss/components/optional/_cookie-warning.scss
```diff
@@ -29,6 +29,7 @@
 	}
 
 	.c-cookieWarning p {
+        @include font-size(small, false);
```
dandel10n: Indentation

jinwood: Done

ashleynolan:
Super NIT – Can you make sure to include a title with the version number.  Makes reading old commits a little more explanatory.

PR #202, opened 2019-02-06
v1.29.0 small styles tweaks for c-orderCard and c-overflowCarousel

dandel10n:
### Changed
- Fix some styles for `c-orderCard` and `c-overflowCarousel` components

### Added
- `.o-tile-link` class

### DIFF
--- a/src/scss/objects/_links.scss
+++ b/src/scss/objects/_links.scss
@@ -25,3 +25,10 @@ a {
     //     text-decoration: none;
     // }
 }
+
+.o-tile-link {
+    text-decoration: none;
+    width: 100%;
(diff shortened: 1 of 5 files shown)


src/scss/objects/_links.scss
```diff
@@ -25,3 +25,10 @@ a {
     //     text-decoration: none;
     // }
 }
+
+.o-tile-link {
```
ashleynolan: This should likely be named something like `o-link--tile` or something similar.

Have you got an example of how it's being used?

dandel10n: I've added changes as discussed

PR #204, opened 2019-02-08
v1.30.0 - Rewrite c-cuisinesWidget to use css grid instead of flex

dandel10n:
### Changed
- Rewrite `c-cuisinesWidget` to use css grid instead of flex

The block looks the same as with flexboxes, just fixed the Edge width calculation problem and reduce a little bit amount of styles

[Gif from Gyazo](https://gyazo.com/c0df29a69863fb93c97fe8c405c2000e)

## Browsers Tested

- [x] Chrome
- [x] Edge
- [x] Internet Explorer 11
- [x] Firefox
- [x] Safari
- [x] Mobile (via browser simulator)

### DIFF
--- a/src/scss/components/optional/_cuisines-widget.scss
+++ b/src/scss/components/optional/_cuisines-widget.scss
@@ -13,6 +13,7 @@
 
 //TODO: Improvement of this solution: WH-682 as suggested in https://github.com/justeat/fozzie/pull/188#issuecomment-442138750
 
+
 $cuisinesWidget-titleColour: $white;
 $cuisinesWidget-defaultBackground: $brand--red;
 $cuisinesWidget-boxShadow: 0 2px 4px 0 rgba(0, 0, 0, 0.1);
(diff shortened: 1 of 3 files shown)


sonnyprince:
@dandel10n should the new css grid stuff also be behind a modifier like it previously was? `.c-cuisinesWidget--gridOfThree `?

Also I think a lot of the PR template thats been removed, especially around testing, would still have a good place for this PR

NIT: A PR title would also be good 😃

dandel10n:
@sonnyprince thanks! Changes now are ready to be reviewed

src/scss/components/optional/_cuisines-widget.scss
```diff
@@ -23,113 +24,78 @@ $cuisinesWidget-gradient: linear-gradient(to top, rgba(0, 0, 0, 0.6), rgba(0, 0,
...
+            }
+        }
+
+        .c-cuisinesWidget-img {
+            vertical-align: bottom;
+            width: 100%;
+        }
+
+        /* autoprefixer grid: autoplace */
+        .c-cuisinesWidget-grid {
```
ashleynolan: Sorry I'm late to the game reviewing this one.  The naming looks a bit off here potentially looking at how the components used.

Dashes are for showing child elements, but it looks like this is the base component.  If so, it's probably better to name this `.c-cuisineGrid`, as `.c-cuisinesWidget-grid` is implying that it's a child of `.c-cuisinesWidget` (which I don't think it is looking at the docs site?).

dandel10n: I've made a PR for this change: https://github.com/justeat/fozzie/pull/206

PR #205, opened 2019-02-11
v1.30.1 - Add pointer cursor for cookie banner close button

xander-marjoram:
The cursor type for the button to close the cookie banner did not change on mouseover, but it should be `pointer` to "feel right" to users.

## UI Review Checks

- [ ] UI Documentation has been updated

## Browsers Tested

- [x] Chrome
- [x] Edge
- [x] Opera

### DIFF
--- a/src/scss/components/optional/_cookie-warning.scss
+++ b/src/scss/components/optional/_cookie-warning.scss
@@ -45,12 +45,12 @@
 		background: url('//dy3erx8o0a6nh.cloudfront.net/images/icon-close-banner.png') no-repeat center center;
 		background-size: 10px 10px;
 		border: none;
-		cursor: pointer;
 
 		&:hover,
 		&:active,
 		&:focus {
 			background-color: $grey--mid;
+			cursor: pointer;
 		}
 	}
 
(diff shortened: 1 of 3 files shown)


src/scss/components/optional/_cookie-warning.scss
```diff
@@ -45,6 +45,7 @@
 		background: url('//dy3erx8o0a6nh.cloudfront.net/images/icon-close-banner.png') no-repeat center center;
 		background-size: 10px 10px;
 		border: none;
+		cursor: pointer;
```
DamianMullins: Should this go on the `: hover` selector below?

xander-marjoram: As cursor only applies on-hover they're functionally equivalent, but I can see your argument from a stylistic point of view

ashleynolan: Agree with this one – should be applied on `:hover` (as would ensure no collisions should styles ever get applied in other states (like `:disabled`).

May not be an issue, but would def make sure it isn't.

PR #207, opened 2019-02-14
v1.32.0 l-container breakpoints and c-pageBanner styles fixes

dandel10n:
### Changed
- `l-container` media queries breakpoints
- `c-pageBanner` small styles amends

### DIFF
--- a/src/scss/base/_layout.scss
+++ b/src/scss/base/_layout.scss
@@ -24,12 +24,12 @@ body {
     padding-left: #{$layout-margin}px;
     padding-right: #{$layout-margin}px;
 
-    @include media('<=mid') {
+    @include media('<mid') {
         padding-left: #{$layout-margin--mid}px;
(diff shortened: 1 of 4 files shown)


src/scss/base/_layout.scss
```diff
@@ -24,12 +24,12 @@ body {
     padding-left: #{$layout-margin}px;
     padding-right: #{$layout-margin}px;
 
-    @include media('<=mid') {
+    @include media('<mid') {
```
dandel10n: styles will be added from <768px

ashleynolan:
@dandel10n Small note for this one – probably worth any of the other verticals currently using these breakpoints (@bulutcy & @sonnyprince & @oleksiishevchukje) To check that this doesn't affect their styling breakpoints when they update to this version of fozzie.

May need some small amends to breakpoint definitions to get them inline (hopefully won't be affected, but just worth keeping an eye on it when you upgrade)

PR #210, opened 2019-03-05
v1.34.0 Small style amends to c-cuisinesGrid--gridOfThree

dandel10n:
### Changed
- Small style amends to `c-cuisinesGrid--gridOfThree`

Replaced grid-gap with margins for better handling less than 6 items grid because if IE explicit grid rule

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 v1.34.0
 ------------------------------
-*March 05, 2019*
+*March 5, 2019*
 
 ### Changed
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -3,6 +3,13 @@
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html)
 
+v1.34.0
+------------------------------
+*March 05, 2019*
```
ashleynolan: Super tiny thing – can just make this `5` rather than `05` as that's consistent with the rest of the single digit dates in the changelog

PR #211, opened 2019-03-06
v1.35.0 Styles for c-orderCard--noOpenRestaurants

dandel10n:
screen shot 2019-03-06 at 15 39 16

### DIFF
--- a/src/scss/components/optional/_order-card.scss
+++ b/src/scss/components/optional/_order-card.scss
@@ -41,6 +41,20 @@ $previousOrder-hoverBoxShadow: 0 2px 8px 0 rgba(0, 0, 0, 0.16), -2px -2px 4px 0 
         }
     }
 
+    .c-orderCard--noOpenRestaurants {
+        min-height: 220px;
+        margin: auto;
+
+        @include media('>=mid') {
+            min-height: 254px;
+        }
+
+        &:hover,
+        &:focus {
+            box-shadow: $previousOrder-boxShadow;
+        }
+    }
+
     .c-orderCard-image {
         width: 100%;
         height: 103px;
@@ -65,6 +79,31 @@ $previousOrder-hoverBoxShadow: 0 2px 8px 0 rgba(0, 0, 0, 0.16), -2px -2px 4px 0 
             bottom: -1px;
             left: 0;
             z-index: zIndex(low);
+        }
+    }
+
+    .c-orderCard-image--noOpenRestaurants {
+        background-color: $grey--lightest;
+    }
+
+    .c-orderCard-iconContainer {
+        width: 64px;
+        height: 64px;
+        border-radius: 50%;
+        background-color: white;
+        position: absolute;
+        top: 50%;
+        right: 50%;
+        transform: translate(50%, -50%);
+
+        svg {
+            width: 40px;
+            height: 40px;
+            position: absolute;
+            top: 50%;
+            right: 50%;
+            transform: translate(50%, -50%);
+            fill: $brand--orange;
         }
     }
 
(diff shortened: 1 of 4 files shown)


src/scss/components/optional/_order-card.scss
```diff
@@ -68,6 +82,31 @@ $previousOrder-hoverBoxShadow: 0 2px 8px 0 rgba(0, 0, 0, 0.16), -2px -2px 4px 0
...
 
+    .c-orderCard-image--noOpenRestaurants {
+        background-color: $grey--lightest;
+    }
+
+    .c-orderCard-iconContainer {
+        width: 64px;
+        height: 64px;
+        border-radius: 50%;
+        background-color: white;
```
kevinrodrigues: Should we be using the variable `$white` here?

src/scss/components/optional/_order-card.scss
```diff
@@ -41,6 +41,20 @@ $previousOrder-hoverBoxShadow: 0 2px 8px 0 rgba(0, 0, 0, 0.16), -2px -2px 4px 0
         }
     }
 
+    .c-orderCard--noOpenRestaurants {
```
sonnyprince: Maybe we could rename this class to `c-orderCard--inactive` or something similar? Just to keep classes a little more generic 😃

src/scss/components/optional/_order-card.scss
```diff
@@ -68,6 +82,31 @@ $previousOrder-hoverBoxShadow: 0 2px 8px 0 rgba(0, 0, 0, 0.16), -2px -2px 4px 0
...
+    .c-orderCard-image--noOpenRestaurants {
+        background-color: $grey--lightest;
+    }
+
+    .c-orderCard-iconContainer {
+        width: 64px;
+        height: 64px;
+        border-radius: 50%;
+        background-color: white;
+        position: absolute;
```
kevinrodrigues: Minor one but it would be nice to combine some of the repeat styles.

```
.c-orderCard-iconContainer,
svg {
  position: absolute;
  top: 50%;
  right: 50%;
  transform: translate(50%, -50%);
}
```

src/scss/components/optional/_order-card.scss
ashleynolan: Yep – 100%.

It would be good to tie this into the `.c-orderCard-image` reference too – so have two options:

```scss
.c-orderCard--inactive {
   .c-orderCard-image {
     …
   }
}

// or could make it a state class so that:
.is-inactive {
	&.c-orderCard {
	   …
    }
    .c-orderCard-image {
      …
    }
}
```

Up to you which one you use really – just a matter of opinion over using a modifier or a state class.

src/scss/components/optional/_order-card.scss
dandel10n: I moved them to utility class instead

dandel10n:
@ashleynolan @kevinrodrigues @sonnyprince thanks, changes are ready for another look

src/scss/trumps/_utilities.scss
```diff
@@ -100,6 +100,13 @@
     top: spacing(x2);
 }
 
+.u-absolutelyCentered {
```
kevinrodrigues: Is it not worth having this as a placeholder too? So you can use it as an extend as well as inline?

dandel10n: It makes sense, but then we need to add it for all utility classes, don't we? I can do it in separate PR

kevinrodrigues: I mean't just for this one as I can see it being handy in other places where others might prefer to use extends instead of adding additional classes to the dom. It's not a biggy tho so happy as it is.

PR #212, opened 2019-03-18
v1.36.0 — Adding new components & utilities.

DamianMullins:
### Added
- Modal optional component.
- Expandable menu modifier.
- Overlay & prevent scroll utility helpers.

### Fixed
- Menu component `has-icon` class only applied to small device widths.

### DIFF
--- a/src/scss/components/optional/_modal.scss
+++ b/src/scss/components/optional/_modal.scss
@@ -1,0 +1,91 @@
+/**
+ * Components > Item Selector
+ * =================================
+ * Used on menu pages to display a modal to the user.
+ *
+ * Example: Menu – When selecting an item.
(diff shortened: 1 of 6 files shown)


src/scss/components/optional/_modal.scss
```diff
@@ -0,0 +1,91 @@
+/**
+ * Components > Item Selector
+ * =================================
+ * Used on menu pages to display a modal to the user. Styled as a normal modal on
```
dandel10n: "Styled as a normal modal on " should there be something else to finish the sentence? :)

DamianMullins: Was a mistake, have removed now.

PR #215, opened 2019-03-21
v1.39.0 c-order-card styles amends

dandel10n:
### Changed
`c-order-card` style amends:

- Restaurant title font-weight should be 500;
- More spacing after order date;
- Order summary text colour should be the same as restaurant title (#333);
- If there is only one line of order summary in one tile, total amount should still be on lower line than summary, height of this part should be taken from the tallest order summary tile;
- "Order again" part should be always at the bottom, should be always the same height (48px);

`c-cuisinesWidget` underline text on hover state

Before:
Screen Shot 2019-03-21 at 11 56 30

After:
Screen Shot 2019-03-21 at 11 28 52

## Browsers Tested

- [x] Chrome
- [x] Safari
- [x] Firefox
- [x] Edge
- [x] Internet Explorer 11
- [ ] Mobile (i.e. iPhone/Android - please list device)

### DIFF
--- a/src/scss/components/optional/_order-card.scss
+++ b/src/scss/components/optional/_order-card.scss
@@ -16,13 +16,13 @@ $previousOrder-hoverBoxShadow: 0 2px 8px 0 rgba(0, 0, 0, 0.16), -2px -2px 4px 0 
 
 @mixin orderCard() {
     .c-orderCard {
-        display: block;
+        display: flex;
+        flex-direction: column;
         border-radius: 8px;
         box-shadow: $previousOrder-boxShadow;
         background-color: $white;
         overflow: hidden;
         text-decoration: none;
-        color: $grey--mid;
         width: 280px;
         max-width: 280px;
         min-height: 259px;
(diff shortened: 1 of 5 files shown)


src/scss/components/optional/_order-card.scss
```diff
@@ -101,18 +101,20 @@ $previousOrder-hoverBoxShadow: 0 2px 8px 0 rgba(0, 0, 0, 0.16), -2px -2px 4px 0
...
         font-weight: $font-weight-bold;
         border-top: 1px solid $grey--offWhite;
         line-height: line-height(16px, 24px);
+        @include font-size(base--scaleUp, false);
+        margin-top: spacing();
     }
 
     .c-orderCard-title {
         font-family: $font-family-base;
+        font-weight: 500;
```
ashleynolan: Should be able to use the font-weight variables for this

dandel10n: Changed :+1:

xander-marjoram:
We should probably test this in IE/Edge too. And on mobile we *probably* don't need the cards to all be the same height (when they're in a single column).

dandel10n:
> We should probably test this in IE/Edge too

@xander-marjoram there is a problem with IE again...I will test once we found out why client side doesn't render again (gtm js error).

dandel10n:
@ashleynolan @xander-marjoram I've tested in IE and Edge :+1:

PR #216, opened 2019-04-03
v1.40.0 — Modal style fix and npm updates.

DamianMullins:
### Changed
- Updated npm dependencies.

### Fixed
- Modal styles which caused the modal to be hidden on large screens.

### DIFF
--- a/src/scss/components/optional/_modal.scss
+++ b/src/scss/components/optional/_modal.scss
@@ -34,6 +34,7 @@ $modal-transitionTime    : 150ms !default;
             background-color: $white;
             border-radius: 8px;
             box-shadow: 0 3px 4px 0 rgba(0, 0, 0, 0.12);
+            display: none;
             padding: spacing(x3);
             position: fixed;
             right: 50%;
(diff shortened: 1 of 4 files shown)


src/scss/components/optional/_modal.scss
```diff
@@ -34,6 +34,7 @@ $modal-transitionTime    : 150ms !default;
             background-color: $white;
             border-radius: 8px;
             box-shadow: 0 3px 4px 0 rgba(0, 0, 0, 0.12);
+            display: none;
```
ashleynolan: Late to the 🎉 …

How come this needs to be `display: none;`? Seems a bit strange to make the content invisible by default.

DamianMullins: It's because the element that `c-modal-content` is applied to is the modal that you see, `c-modal` is the wrapper which holds the content.

Looking back at the code now it does look like it can be improved, I have some other styles I need to add so I can look at fixing the modal styles up whilst I'm at it.

PR #218, opened 2019-04-11
v1.42.0 — Added OK, dark, and indicator badge modifiers.

DamianMullins:
### Added
- OK, dark, and indicator badge modifiers.

### DIFF
--- a/src/scss/components/optional/_badges.scss
+++ b/src/scss/components/optional/_badges.scss
@@ -23,9 +23,15 @@ $badge-important-color      : $white;
 $badge-important-bgColor    : $orange;
 $badge-award-color          : $purple;
 $badge-award-bgColor        : $purple--light;
+$badge-ok-color             : $white;
+$badge-ok-bgColor           : $green;
+$badge-dark-color           : $white;
+$badge-dark-bgColor         : $grey--darkest;
+$badge-indicator-color      : $blue;
+$badge-indicator-bgColor    : $white;
 $badge-padding              : 1px 5px;
-$badge-rounded-padding      : 4px spacing(x2);
 $badge-rounded-radius       : 14px;
+$badge-rounded-padding      : spacing(x0.5) spacing(x2);
 $badge-angled-radius        : 2px;
 
 @mixin badge() {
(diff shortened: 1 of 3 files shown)


src/scss/components/optional/_badges.scss
```diff
@@ -23,9 +23,15 @@ $badge-important-color      : $white;
 $badge-important-bgColor    : $orange;
 $badge-award-color          : $purple;
 $badge-award-bgColor        : $purple--light;
+$badge-ok-color             : $white;
+$badge-ok-bgColor           : $green;
+$badge-dark-color           : $white;
```
kevinrodrigues: Do you need to specify this one? Line 26 could be used instead?

DamianMullins: but that variable is for the OK button?

kevinrodrigues: OK. So a more generic one would be better so we don't end up with loads of variables pointing to the same color. but it's fine.

ashleynolan: I'm for this as long as we don't end up with hundreds of badge types in the future (consolidation then would be good). I guess if that happens though we can optimise it then as there's still only a handful of types.

src/scss/components/optional/_badges.scss
```diff
@@ -77,6 +83,21 @@ $badge-angled-radius        : 2px;
             color: $badge-award-color;
         }
 
+        .c-badge--ok {
```
ashleynolan: May be worth adding a comment as to what this modifier is – `ok` is a bit vague as to the type (compared to `error` etc).  Is it used in combination with messsages?

DamianMullins: It's only used on the item quantity indicator on menus, see attached image. I wasn't really sure what else to call it but as it's green I went with OK 😸 

Screenshot 2019-04-11 at 12 31 21

DamianMullins: Any thoughts on what else it could be called?

DamianMullins: I have renamed to `c-badge--success--alt` and added a description.

ashleynolan: Sorry for delay in replying – I was away Thurs/Fri last week

That name looks good though 👍

PR #223, opened 2019-05-09
v1.47.0 - Add user message component

xander-marjoram:
Styles for user message component.

## UI Review Checks
- [ ] This PR has been checked with regard to our brand guidelines
- [ ] UI Documentation has been [created|updated]

## Browsers Tested
- [x] Chrome
- [x] Firefox
- [x] Edge
- [ ] Internet Explorer 11
- [x] Mobile emulator (Chrome)

### DIFF
--- a/src/scss/components/optional/_user-message.scss
+++ b/src/scss/components/optional/_user-message.scss
@@ -1,0 +1,53 @@
+/**
+ * User Message Component
+ * =================================
+ * Used to display the user message banner on the page.
+ *
+ * *The `c-userMessage` component is an optional mixin within Fozzie — if you'd like to use it in your project you can include it by adding `@include userMessage();` into your SCSS dependencies file.*
+ */
+
+@mixin userMessage() {
+    .c-userMessage {
+        color: #fff;
+        background-color: $orange;
+        max-width: 100%;
+        margin-top: spacing(x2);
+    }
+
+    .c-userMessageContainer {
+        max-width: 350px;
+        margin: 0 auto;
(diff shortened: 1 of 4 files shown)


src/scss/components/optional/_user-message.scss
```diff
@@ -0,0 +1,53 @@
...
+            fill: white;
+            width: 28px;
+            height: 28px;
+            display: block;
+            align-self: center;
+        }
+    }
+
+    .c-userMessageText {
+        @include font-size(base);
```
ashleynolan: Does this need to be set because the text isn't inside a `p` tag? (just interested as can't see the markup here 😄 )

src/scss/components/optional/_user-message.scss
```diff
@@ -0,0 +1,53 @@
...
+        grid-template-columns: 40px auto;
+
+        @include media('>=narrow') {
+            max-width: 540px;
+        }
+
+        @include media('>=mid') {
+            max-width: 650px;
+            margin: 0 auto;
+            padding: spacing(x2) 0 spacing(x2) 36px;
```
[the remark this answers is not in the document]
xander-marjoram: I don't think this padding is needed at all actually, I think it was to help centre the text when we had the SVG in the `:before`.

src/scss/components/optional/_user-message.scss
xander-marjoram: It looks like it doesn't need to be set, because it is in a `p` tag 👍

PR #227, opened 2019-05-31
v1.50.0 - Usage of the Glaze theme

oleksiishevchukje:


src/scss/settings/_variables.scss
```diff
@@ -75,6 +75,12 @@ $font-weight-headings-preload: 400;
     $font-weight-headings   : 800; // Heading style is Aspira Heavy, which is 800 font-weight
 }
 
+// Glaze font overrides
+// ==========================================================================
+@if ($theme == 'gl') {
+    $font-family-base       : 'Open Sans';
+    $font-family-headings   : 'Nunito';
```
oleksiishevchukje: Not sure about the fallbacks for these font families... is it fine to not have them?

ashleynolan: It'll work without them, just if the fonts don't load you'll end up with default system fonts, so it's best for design to specify which system font they'd rather have as their default (Arial, Times New Roman etc)

oleksiishevchukje: @ashleynolan Thanks. I will clarify this. I would update this separately, if it is fine.

ashleynolan: Yep – that's fine 😄

PR #232, opened 2019-06-18
v1.56.0 - Adding styles for the cuisineCollage.

SachaZvetelman:
This has been reviewed and approved on an internal PR already by @kevinrodrigues and @ashleynolan.

### DIFF
--- a/src/scss/components/optional/_cuisines-collage.scss
+++ b/src/scss/components/optional/_cuisines-collage.scss
@@ -1,0 +1,133 @@
+/**
+* Components > Cuisines Collage
+* =================================
+*
+* Example: Homepage – Popular cuisines collage.
+*
+* The `c-cuisinesCollage` component is an optional mixin within Fozzie.
+* If you'd like to use it in your project you can include it by adding `@include cuisinesCollage();` into your SCSS dependencies file.
+*
+* Documentation:
+* https://fozzie.just-eat.com/styleguide/ui-components/cuisines-collage
+*/
+
+
+@mixin cuisinesCollage() {
+    @include media-context(('narrow-mid': 570px, 'mid-wide': 960px, 'wide-huge': 1240px)) {
+        .c-cuisinesCollage-items {
+            width: 100%;
+            height: 330px;
+            display: flex;
+            flex-flow: row nowrap;
+            margin-bottom: spacing(x2);
+            justify-content: center;
+            align-items: flex-start;
+            align-content: normal;
+
+            button {
+                display: block;
+                width: 100%;
+            }
+
+            img {
+                border-radius: $border-radius;
+                object-fit: cover;
+            }
+        }
+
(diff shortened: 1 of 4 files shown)


sonnyprince:
@SachaZvetelman is this much different to `c-cuisinesWidget`?

sonnyprince:
would also be good to have screenshots etc 😃

SachaZvetelman:
@sonnyprince, I guess that's quite subjective 😄. Here's the screenshot for UK:

image

dandel10n:
@SachaZvetelman you need to include the `_cuisines-collage.scss` file to fozzie/src/scss/_dependencies.scss

src/scss/components/optional/_cuisines-collage.scss
```diff
@@ -0,0 +1,133 @@
...
+            width: 100%;
+            height: 330px;
+            display: flex;
+            flex-flow: row nowrap;
+            margin-bottom: spacing(x2);
+            justify-content: center;
+            align-items: flex-start;
+            align-content: normal;
+
+            button {
```
sonnyprince: could we target by class here, maybe `c-cuisinesCollage-button`?

src/scss/components/optional/_cuisines-collage.scss
```diff
@@ -0,0 +1,133 @@
...
+                width: 100%;
+                max-width: 461px;
+            }
+
+            button:first-child img {
+                margin-bottom: 5px;
+            }
+        }
+
+        @include media('<wide-huge') {
```
sonnyprince: Is it better to have this nested within line 65? Keeps an element and all responsive styles in one place.

src/scss/components/optional/_cuisines-collage.scss
```diff
@@ -0,0 +1,133 @@
...
+            .c-cuisinesCollage-item:nth-child(1) {
+                display: none;
+            }
+        }
+
+        @include media('<mid-wide') {
```
sonnyprince: Same comment as above - better to keep styles together?

src/scss/components/optional/_cuisines-collage.scss
```diff
@@ -0,0 +1,133 @@
...
+        .c-cuisinesCollage-item:last-child {
+            margin-right: 0;
+
+            img {
+                height: 161px;
+                width: 100%;
+                max-width: 461px;
+            }
+
+            button:first-child img {
```
sonnyprince: Is there any other way to target this? quite specific

src/scss/components/optional/_cuisines-collage.scss
sonnyprince: There's a few places throughout that may be better to re-shuffle and have the responsive styles live closer to the elements they are changing 🤔

PR #244, opened 2019-10-01
v1.66.0 Improved CSS fallback font cascade

markedup:
Simple update to the CSS font fallback cascade using more visually harmonious fallbacks, e.g. sans-serif system fallbacks for all sans-serif Web fonts (likewise serif-for-serif).

## UI Review Checks

## Browsers Tested
- [x] Chrome
- [ ] Edge
- [ ] Internet Explorer 11
- [ ] Mobile (i.e. iPhone/Android - please list device)

### DIFF
--- a/src/scss/base/_typography.scss
+++ b/src/scss/base/_typography.scss
@@ -28,6 +28,7 @@
         font-family: Aspira;
         font-style: normal;
         font-weight: 800;
+        font-display: swap;
         src: url('#{$font-path}/menulog/aspira-heavy.woff2') format('woff2'),
              url('#{$font-path}/menulog/aspira-heavy.woff') format('woff'),
              url('#{$font-path}/menulog/aspira-heavy.ttf') format('truetype');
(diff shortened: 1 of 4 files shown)


src/scss/base/_typography.scss
```diff
@@ -28,6 +28,7 @@
         font-family: Aspira;
         font-style: normal;
         font-weight: 800;
+        font-display: swap;
         src: url('#{$font-path}/menulog/aspira-heavy.woff2') format('woff2'),
              url('#{$font-path}/menulog/aspira-heavy.woff') format('woff'),
              url('#{$font-path}/menulog/aspira-heavy.ttf') format('truetype');
```
kevinrodrigues: I don't think we need ttf support now.

markedup: @kevinrodrigues I wondered about that too. It's not in scope for this ticket, but can remove the path reference to it if you wish?

kevinrodrigues: Not a biggy here tbh but it would be good to remove them in another ticket if they're not required anymore.

PR #245, opened 2019-10-07
v1.67.0 - update unit intervals to fix safari rounding issue

sonnyprince:
- update unit intervals to fix safari rounding issue

## Browsers Tested

- [x] Safari
- [x] Edge
- [x] Internet Explorer 11
- [x] Safari Mobile

### DIFF
--- a/src/scss/settings/_variables.scss
+++ b/src/scss/settings/_variables.scss
@@ -83,8 +83,16 @@ $font-weight-headings-preload: 400;
 }
 
 
-//  Global Breakpoints
+//  Global Breakpoints and unit intervals
 // ==========================================================================
+
+$unit-intervals: (
+  'px': 1,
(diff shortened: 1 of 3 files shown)


src/scss/settings/_variables.scss
```diff
@@ -83,8 +83,16 @@ $font-weight-headings-preload: 400;
 }
 
 
-//  Global Breakpoints
+//  Global Breakpoints and unit intervals
 // ==========================================================================
+
+$unit-intervals: (
+  'px': 1,
+  'em': 0.063, //0.063ems converts to 1px. We need this as some browsers are very specific with breakpoints
```
ashleynolan: Probably worth putting the calculation in the comments (as this is based on a base font-size of `16px`).  So `1 / 16 = 0.0625`

PR #249, opened 2019-10-30
v1.71.0 -- Updating mediaElement component

smilburn1990:
### Changed
- Styles for `.c-mediaElement-infoLink`, `$mediaElement-img-borderRadius` and `.c-mediaElement--fullstack--negativeTop` for menuweb UI designs

Screenshot 2019-10-30 at 11 45 04
Screenshot 2019-10-30 at 11 44 45

### DIFF
--- a/src/scss/components/_media-element.scss
+++ b/src/scss/components/_media-element.scss
@@ -126,6 +126,11 @@ $mediaElement-spacing           : spacing(x2);
             margin-top: spacing(x0.5);
         }
 
+        .c-mediaElement--fullstack--negativeTop & {
+            flex: auto;
+            margin-top: -#{spacing(x2)} !important;
+        }
+
         // some content uses br for content shaping – get rid of it on mobile devices
         & br {
             @include media('<mid') {
--- a/src/scss/components/_media-element.scss
+++ b/src/scss/components/_media-element.scss
@@ -123,7 +123,6 @@ $mediaElement-spacing           : spacing(x2);
 
         .c-mediaElement--fullstack & {
             flex: auto;
-            margin-top: (spacing(x2) * -1);
         }
 
         // some content uses br for content shaping – get rid of it on mobile devices
--- a/src/scss/components/_media-element.scss
+++ b/src/scss/components/_media-element.scss
@@ -123,6 +123,7 @@ $mediaElement-spacing           : spacing(x2);
 
         .c-mediaElement--fullstack & {
             flex: auto;
+            margin-top: spacing(x0.5);
         }
 
         // some content uses br for content shaping – get rid of it on mobile devices
--- a/src/scss/components/_media-element.scss
+++ b/src/scss/components/_media-element.scss
@@ -12,7 +12,7 @@
 $mediaElement-img-width         : 50px;
 $mediaElement-img-width--mid    : 78px;
 $mediaElement-img-width--wide   : 67px;
-$mediaElement-img-borderRadius  : 8px;
+$mediaElement-img-borderRadius  : 2px;
 $mediaElement-infoLinkColor     : $blue;
 $mediaElement-fontSize          : base--scaleUp;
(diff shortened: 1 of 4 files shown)


src/scss/components/_media-element.scss
```diff
@@ -123,7 +123,7 @@ $mediaElement-spacing           : spacing(x2);
 
         .c-mediaElement--fullstack & {
             flex: auto;
-            margin-top: spacing(x0.5);
+            margin-top: (spacing(x2) * -1);
```
DamianMullins: I'd leave this and try applying a class like `u-spacingTop--large` to the element you're looking to style.

src/scss/components/_media-element.scss
```diff
@@ -12,7 +12,7 @@
 $mediaElement-img-width         : 50px;
 $mediaElement-img-width--mid    : 78px;
 $mediaElement-img-width--wide   : 67px;
-$mediaElement-img-borderRadius  : 8px;
+$mediaElement-img-borderRadius  : 2px;
```
DamianMullins: Did this change come from the UX team?

smilburn1990: https://app.zeplin.io/project/5a7af4078b4a016801348b6c/screen/5b72f088e47db8634b32b23c @DamianMullins

src/scss/components/_media-element.scss
smilburn1990: You mean create a new one that is similar @DamianMullins -because I need it to be `-16px`?

DamianMullins: In that can you use the `-#{spacing(x2)}` syntax to keep it consistent with other negative values?

Also, remember that this class isn't just used in Menu Web, it could be in use anywhere that Fozzie is used in a project. So it's probably worth creating a modifier for this style so it doesn't break layouts anywhere else.

DamianMullins: Also, do you want the negative spacing across all device widths?

src/scss/components/_media-element.scss
```diff
@@ -123,7 +123,6 @@ $mediaElement-spacing           : spacing(x2);
 
         .c-mediaElement--fullstack & {
             flex: auto;
-            margin-top: spacing(x0.5);
```
DamianMullins: Can you add this one back in, please.

src/scss/trumps/_utilities.scss
```diff
@@ -212,6 +212,9 @@
 .u-spacingTop {
     margin-top: spacing() !important;
 }
+    .u-spacingTop--minus {
```
DamianMullins: I meant a modifier of the `c-mediaElement--fullstack` class, so something like `c-mediaElement--fullstack--negativeTop` after the `.c-mediaElement--fullstack &` rule?

src/scss/components/_media-element.scss
smilburn1990: Done 👍

src/scss/trumps/_utilities.scss
smilburn1990: Done 👍

src/scss/components/_media-element.scss
ashleynolan: Might be worth confirming with them that this with ripple out wherever there's a logo (like on home and search)

src/scss/components/_media-element.scss
```diff
@@ -126,6 +126,11 @@ $mediaElement-spacing           : spacing(x2);
             margin-top: spacing(x0.5);
         }
 
+        .c-mediaElement--fullstack--negativeTop & {
+            flex: auto;
+            margin-top: -#{spacing(x2)} !important;
```
DamianMullins: Do you need `!important` here? This is declared after the previous rule, where margin was being applied, so the cascade should mean that this syle has precedence?

PR #253, opened 2019-11-26
v1.74.0 - Porting styles from internal projects.

DamianMullins:
### Added
- Modifiers for `c-card` component.

### Changed
- Updated styles for `c-mediaElement` component.
- Updated styles for `c-menu` component.
- Updated styles for `c-modal` component.


src/scss/components/optional/_menu.scss
```diff
@@ -168,17 +122,61 @@ $menu--expandable-linkActiveFontSize    : base--scaleUp !default;
...
+
+                .c-menu-link--active {
+                    @include font-size($menu--expandable-linkActiveFontSize);
+                    text-decoration: underline;
+                    width: 100vw;
+                }
+            }
+        }
+
+        .c-menu--expandable--expanded {
```
benwhite-deltas: Should this be another indent-level in relative to `.c-menu--expandable`, for consistency with the namespace/indentation rules?

DamianMullins: It should, I'll do that in a following PR if that's OK 😸

src/scss/components/optional/_menu.scss
```diff
@@ -18,6 +18,7 @@ $menu-border-width                      : 1px;
 $menu-border-width--active              : 2px;
 $menu-link-color                        : $color-text;
 $menu-link-padding                      : spacing() spacing(x2);
+$menu-positionTop                       : $stickyHeader-height + spacing(x2);
```
ashleynolan: Just pulled in the latest version of fozzie and I'm getting errors around this variable:

`Undefined variable: "$stickyHeader-height".`

Does this get defined anywhere atm?

DamianMullins: It's defined in Menu Web which would explain why we didn't see the error when testing, sorry! The value is always 57px. Do you want a fix PR?

ashleynolan: Yes please 👍

PR #262, opened 2019-12-12
v1.79.0: Changed styles for menuweb

smilburn1990:
### Changed
- Styles for `c-mediaElement-img--negativeTop` and `c-mediaElement-heading`


src/scss/components/_media-element.scss
```diff
@@ -111,7 +111,7 @@ $mediaElement-spacing           : spacing(x2);
     // positions the image over the top edge of the media element
     .c-mediaElement-img--negativeTop {
         margin-top: 0;
-        position: absolute;
```
DamianMullins: Because this is now relative, the position of the image is slightly off. I changed them to be absolute recently so that we had consistent positioning across all breakpoints.

PR #279, opened 2020-01-20
v1.82.0 - Component updates.

DamianMullins:
### Added
- `c-badge` component small rounded modifier styles.
- `u-spacingBottom--large--aboveMid` & `u-shadowBottom--belowMid` helper to the utilities.

### Changed
- Updated `c-card` component highlight modifier styles.
- Updated `c-modal` component styles.
- Consistent spacing between rules in the utilities module.

### Removed
- `c-card` component large rounded modifier styles.

### DIFF
--- a/src/scss/components/_cards.scss
+++ b/src/scss/components/_cards.scss
@@ -12,8 +12,7 @@ $card-bgColor--disabled                         : $grey--lightest;
 $card-tooltip-width                             : 10px;
 $card-arrow-bottom-position                     : 0;
 $card-padding                                   : spacing(x2);
-$card-radius                                    : 4px;
-$card-radius--large                             : 8px;
+$card-radius                                    : spacing(x0.5);
 $card-borderColor                               : $grey--lightest;
 $card-section-margin                            : spacing(x4);
 $card-section-margin--large                     : spacing(x8);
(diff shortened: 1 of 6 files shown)


src/scss/components/_cards.scss
```diff
@@ -12,8 +12,7 @@ $card-bgColor--disabled                         : $grey--lightest;
 $card-tooltip-width                             : 10px;
 $card-arrow-bottom-position                     : 0;
 $card-padding                                   : spacing(x2);
-$card-radius                                    : 4px;
-$card-radius--large                             : 8px;
+$card-radius                                    : spacing(x0.5);
```
ashleynolan: Not a big deal as it's unlikely to change atm, but not sure we'd want spacing linked to the border-radius like this. Would be weird if the spacing increased that corners of components also were affected.

DamianMullins: Yeah, good shout, I get a bit carried away with the spacing helper sometimes! I'll update this soon when I next put a PR in.

DamianMullins: This was done in #280.

DamianMullins: Ah no it wasn't, I updated the wrong thing!

DamianMullins: Fixed in #281 😸

PR #280, opened 2020-01-21
v1.83.0 - Revert max height change to fix scrolling issue.

DamianMullins:
### Fixed
- Reverted `max-height` change to `.c-modal-content-scrollable` as it prevents scrolling on small screen devices.

### Changed
- Use explicit value for rounded badge border radius.

### DIFF
--- a/src/scss/components/optional/_modal.scss
+++ b/src/scss/components/optional/_modal.scss
@@ -139,6 +139,7 @@ $modal-borderRadius     : 4px;
             }
 
         .c-modal-content-scrollable {
+            max-height: 100vh;
             overflow-x: hidden;
             overflow-y: auto;
             position: relative;
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,9 @@ v1.83.0
 ### Fixed
 - Reverted `max-height` change to `.c-modal-content-scrollable` as it prevents scrolling on small screen devices.
 
+### Changed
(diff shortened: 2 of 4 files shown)


src/scss/components/optional/_modal.scss
```diff
@@ -139,6 +139,7 @@ $modal-borderRadius     : 4px;
             }
 
         .c-modal-content-scrollable {
+            max-height: 100vh;
```
asherstoppard: Have we checked how this looks visually on iOS devices as `100vh` extends below the bottom bar? It might be worth looking at `max-height: -webkit-fill-available;` with the appropriate `100vh` fallback.

Reference: [CSS3 100vh not constant in mobile browser](https://stackoverflow.com/questions/37112218/css3-100vh-not-constant-in-mobile-browser).

DamianMullins: Will test that out in another PR, this one is to revert a change which has broken a consumer of this style and is blocking a release.

DamianMullins: We've hit all sorts of issues with mobile heights on different devices and OSs due in most part to the browser UI which they add in, such as the nav bar at the bottom in iOS, that doesn't respect the 100vh values correctly.

asherstoppard: Yeah it's a royal pain. Sounds good though, thanks for clarifying :)

PR #283, opened 2020-01-23
v1.85.0 - Update overflow-carousel

InsungMulumba:
## Before
image

## After
image

## Browsers Tested

- [x] Chrome
- [x] Edge
- [x] Internet Explorer 11
- [x] Firefox
- [x] Mobile (i.e. iPhone/Android - please list device)

### DIFF
--- a/src/scss/components/optional/_overflow-carousel.scss
+++ b/src/scss/components/optional/_overflow-carousel.scss
@@ -94,6 +94,17 @@
             }
         }
     }
+
+    .c-overflowCarousel--hideScroll {
+        @include media('>=wide') {
+            scrollbar-width: none;
+        }
+    }
+
+    .c-overflowCarousel--hideScroll::-webkit-scrollbar {
+        width: 0px;
+        display: none;
+    }
 }
 
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,14 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html)
 
 
+v1.85.0
+------------------------------
+*January 23, 2020*
+
+### Changed
+- Updated overflow carousel to hide scrollbar on YourFavourites component
+
+
 v1.84.0
 ------------------------------
 *January 22, 2020*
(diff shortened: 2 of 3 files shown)


InsungMulumba:
https://streamable.com/m1kfc

Should not affect ability to scroll.

src/scss/components/optional/_overflow-carousel.scss
```diff
@@ -94,6 +94,17 @@
             }
         }
     }
+
+    .c-overflowCarousel--hideScroll {
+        @include media('>=wide') {
+            scrollbar-width: none;
+        }
```
InsungMulumba: For Firefox

src/scss/components/optional/_overflow-carousel.scss
```diff
@@ -94,6 +94,17 @@
             }
         }
     }
+
+    .c-overflowCarousel--hideScroll {
```
kevinrodrigues: I don't think we should be hiding the default scrollbar IMO, was there a valid reason for removing it at all? I wouldn't be surprised if there are lots of users that manually drag the scrollbar using the mouse. Also, it gives a good indication to the user that there are more items available within the carousel.

InsungMulumba: The reason I was given is that it is an 'eyesore' and doesn't fit the styling especially in comparison with the other carousels.

The carousel is going to be limited to only 3 items at a time, so users on desktop should be able to see all 3 at once when the screen is rendered, or be able to scroll (as shown in the video)
I agree that users may not intuitively know that there are more items on the screen and I raised this before developing, the UX designer and product manager on our team made this decision.

ashleynolan: I tend to agree (would be interesting to know the accessibility implications if there are any, as not sure if AA standards mention scrollbar usage like this?)

In terms of it being an eyesore – I agree it def wasn't ideal before. There are properties available that could improve the look of the scrollbar ([at least in webkit](https://css-tricks.com/the-current-state-of-styling-scrollbars/)). Not sure if this could be considered potentially as an alternative.

kevinrodrigues: The other option might be to make the 3 cards a little more responsive and then make it trigger the overflow hidden from the mid-break point instead, that way on smaller screens you can scroll correctly and on desktops it's fully visible?

InsungMulumba: @ashleynolan @kevinrodrigues Can I set up a quick meeting for the two of you to come and discuss this with Conversion tomorrow?

PR #285, opened 2020-01-28
v1.86.0: New class for the scroll height

smilburn1990:
### Added
- New class `c-modal-content-scrollHeight` for the scroll height of the mega modal

### DIFF
--- a/src/scss/components/optional/_modal.scss
+++ b/src/scss/components/optional/_modal.scss
@@ -138,11 +138,8 @@ $modal-borderRadius     : 4px;
                 display: block;
             }
 
-        .c-modal-content-scrollable {
+        .c-modal-content-scrollHeight {
             max-height: 100vh;
-            overflow-x: hidden;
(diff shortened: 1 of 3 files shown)


src/scss/components/optional/_modal.scss
```diff
@@ -157,6 +154,12 @@ $modal-borderRadius     : 4px;
             }
         }
 
+        .c-modal-content-scrollable {
```
zlatinivanov-je: Is this missing media queries?

smilburn1990: Nope, if it is scrollable the class will be applied on the vertical via logic 👍

smilburn1990: We basically have a bug on menuweb where we want to turn the `max-height` off for one of many modal instances. So the best way is to separate out the scrollability and the heights themselves if that makes sense?

PR #287, opened 2020-02-04
v1.88.0: New helper to _rwd.scss for hiding items on tiny devices

InsungMulumba:
# Changelog
### Added 
New helper to `_rwd.scss` for hiding items on tiny devices

[Relevant Pull Request with zeplin attached](https://jira.just-eat.net/browse/CCS-218)

The page banner picture needs to be hidden on 375px  (and lower) devices

### DIFF
--- a/src/scss/components/optional/_overflow-carousel.scss
+++ b/src/scss/components/optional/_overflow-carousel.scss
@@ -97,6 +97,7 @@
 
     .c-overflowCarousel--hideScroll {
         @include media('>=wide') {
+         /* stylelint-disable-next-line property-no-unknown */
             scrollbar-width: none;
         }
     }
--- a/src/scss/components/optional/_overflow-carousel.scss
+++ b/src/scss/components/optional/_overflow-carousel.scss
@@ -97,7 +97,7 @@
 
     .c-overflowCarousel--hideScroll {
         @include media('>=wide') {
-         /* stylelint-disable-next-line property-no-unknown */
+            /* stylelint-disable-next-line property-no-unknown */
             scrollbar-width: none;
         }
     }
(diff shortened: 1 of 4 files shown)


src/scss/trumps/_rwd.scss
```diff
@@ -36,6 +36,12 @@
     }
 }
 
+.u-showAboveTiny {
+    @include media('<=tiny') {
```
InsungMulumba: Is it ok to use `<=` rather than `<` here? This work is for 375px devices and smaller

src/scss/components/optional/_overflow-carousel.scss
```diff
@@ -97,6 +97,7 @@
 
     .c-overflowCarousel--hideScroll {
         @include media('>=wide') {
+         /* stylelint-disable-next-line property-no-unknown */
```
InsungMulumba: Is this indentation ok?

src/scss/trumps/_rwd.scss
xander-marjoram: I wonder if `u-hideBelowTiny` might better describe what this class does?

src/scss/components/optional/_overflow-carousel.scss
xander-marjoram: If you can fix it, I probably would. And while you're at it could you bump the date in the changelog, please!

InsungMulumba: I was more asking about what the best practice is, should the comment line up with the @include?

src/scss/trumps/_rwd.scss
InsungMulumba: That was my initial thought, but I thought to follow the standards of the other classes, @xander-marjoram I'm happy to rename if you strongly recommend

PR #288, opened 2020-02-06
v1.89.0 - Updating modal styles & npm dependencies.

DamianMullins:
### Added
- Reduce motion query to `c-modal--popUp` which disables transition if user has setting enabled.
- Modal close button styles.

### Changed
- `c-modal--popUp--belowMid` modifier changed to `c-modal--popUp`.
- Switch `js-test-budy` for `@justeat/js-test-buddy`.
- Updated npm dependencies.

### Removed
- Unused `c-modal-actions--raised` & `c-modal-actions--fixed` classes.

### Fixed
- `c-modal--popUp` now applies transition correctly.
- Ignore test files when transpiling with Babel.

### DIFF
--- a/package.json
+++ b/package.json
@@ -69,7 +69,7 @@
     "lint": "yarn run lint:css && yarn run lint:js",
     "lint:css": "gulp scss:lint --prod",
     "lint:js": "gulp scripts:lint --prod",
-    "prepareOnly": "concurrently -n \"lint,compile,test\" -c \"blue,yellow,green\" \"yarn lint\" \"yarn compile\" \"yarn test\" --kill-others-on-fail",
+    "prepare": "concurrently -n \"lint,compile,test\" -c \"blue,yellow,green\" \"yarn lint\" \"yarn compile\" \"yarn test\" --kill-others-on-fail",
     "test": "gulp scripts:test",
     "test:cover": "gulp scripts:test:coverage",
     "test:cover:CI": "cat coverage/lcov.info | coveralls"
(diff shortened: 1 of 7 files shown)


src/scss/components/optional/_modal.scss
```diff
@@ -70,130 +85,156 @@ $modal-borderRadius     : 4px;
...
-                @include media('>=mid') {
-                    padding: spacing(x5) spacing(x7) spacing(x2);
-                }
+        .c-modal-closeBtn--rounded {
+            border: solid 1px $white;
+            background-color: $white;
+            border-radius: 50%;
+            opacity: 0.9;
+
+            &:hover {
```
kevinrodrigues: Do we need a `focus` here too?

DamianMullins: It's only there to override an `o-btn--icon:hover` hover style (which only overrides the default browser hover style). I tested this using focus and active and all is fine.

package.json
```diff
@@ -34,21 +34,21 @@
...
     "fontfaceobserver": "^2.0.13",
     "include-media": "1.4.9",
     "kickoff-grid.css": "1.2.0",
     "normalize-scss": "7.0.1"
   },
   "devDependencies": {
-    "@justeat/gulp-build-fozzie": "^8.4.0",
-    "concurrently": "4.1.0",
-    "coveralls": "3.0.3",
+    "@justeat/gulp-build-fozzie": "^9.0.1",
```
ashleynolan: Can you remove the `^`?

DamianMullins: 💩 done.

PR #289, opened 2020-02-10
v1.90.0: Changed button colour for menulog

smilburn1990:
### Added 
- New colors for button `:hover` and `:focus`

### DIFF
--- a/src/scss/objects/_buttons.scss
+++ b/src/scss/objects/_buttons.scss
@@ -72,6 +72,8 @@ $btn-rounded-width                  : 38px;
     $btn-primary-bgColor--focus           : $green--darkest;
 
     $btn-secondary-textColor              : $green;
+    $btn-secondary-textColor--active      : $green;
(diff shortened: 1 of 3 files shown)


src/scss/objects/_buttons.scss
```diff
@@ -72,6 +72,8 @@ $btn-rounded-width                  : 38px;
     $btn-primary-bgColor--focus           : $green--darkest;
 
     $btn-secondary-textColor              : $green;
+    $btn-secondary-textColor--active      : $green;
+    $btn-secondary-textColor--hover       : $green;
```
kevinrodrigues: Do you need one for `focus` at all?

smilburn1990: Yep, gets used in the modal 👍

kevinrodrigues: ok. Just that the other variables explicitly set a focus style even tho it's the same colour. Also the changelog says added `:hover & :focus`.

PR #293, opened 2020-03-09
v2.1.0 - Updated modal max width.

DamianMullins:
### Fixed
- Updated `c-modal` max-width to an even number as odd-numbered widths cause text to become blurry in some browsers.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,14 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html)
 
 
+v2.1.0
+------------------------------
+*March 10, 2020*
+
+### Fixed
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -4,6 +4,14 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html)
 
 
+v2.1.0
+------------------------------
+*March 10, 2020*
+
+### Fixed
+- Updated `c-modal` max width to an even number as odd numbered widths cause text to become blurry in some browsers.
```
ashleynolan: Is this really a thing, or is it just a case of OCD against odd numbers 🤣

DamianMullins: Haha yes, it has been an issue for a few months now and only tracked this down a few days back. Something about subpixel font rendering where it can't render fonts correctly when it divides a value and the result is not a whole number.

DamianMullins: I do also have an OCD thing about odd numbers though..!

PR #306, opened 2020-07-22
v3.0.0-beta.8 - Updating various uses of grey across components

ashleynolan:
### Changed
- Updating various uses of grey across components and updating `fozzie-colour-palette` version with new alias colours for default text and borders.

## UI Review Checks

## Browsers Tested

- [x] Chrome

### DIFF
--- a/src/scss/components/optional/_fullscreen-pop-over.scss
+++ b/src/scss/components/optional/_fullscreen-pop-over.scss
@@ -15,7 +15,7 @@
 $fullScreenPopOver-background        : $grey--offWhite;
 $fullScreenPopOver-action-background : $white;
 $fullScreenPopOver-padding           : spacing(x2);
-$fullScreenPopOver-border-color      : $grey--lighter;
+$fullScreenPopOver-border-color      : $grey--light;
 $fullScreenPopOver-shadow-color      : rgba($black, 0.12);
 
 @mixin fullScreenPopOver() {
--- a/src/scss/objects/_form-controls.scss
+++ b/src/scss/objects/_form-controls.scss
@@ -91,7 +91,7 @@
          display: inline-block;
          width: $formControl-width;
          height: $formControl-height;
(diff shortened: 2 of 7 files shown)


src/scss/components/optional/_fullscreen-pop-over.scss
```diff
@@ -15,7 +15,7 @@
 $fullScreenPopOver-background        : $grey--offWhite;
 $fullScreenPopOver-action-background : $white;
 $fullScreenPopOver-padding           : spacing(x2);
-$fullScreenPopOver-border-color      : $grey--lighter;
+$fullScreenPopOver-border-color      : $grey--light;
```
xander-marjoram, line 18: Should we be using `$color-border--interactive`?

src/scss/objects/_form-controls.scss
```diff
@@ -91,7 +91,7 @@
          display: inline-block;
          width: $formControl-width;
          height: $formControl-height;
-         border: $formControl-border-width solid $grey--lighter;
+         border: $formControl-border-width solid $grey--light;
```
xander-marjoram, line 94: Should we be using `$color-border--interactive`?

src/scss/components/optional/_fullscreen-pop-over.scss
ashleynolan: I'm not entirely sure tbh (as we don't have designs for these so this is just my best guess against what we have in zeplin).

We'll be switching these variables out for PIE variables soon (when I integrate that in the next couple of weeks) so I'll make sure I go through and update all the aliases where I can when that happens.

src/scss/objects/_form-controls.scss
ashleynolan: Same here – I'm not 100% sure, I think this will have it's own unique token when we switch to PIE aliases.

PR #310, opened 2020-08-21
v4.0.0-beta.0 - Updating type map and font declarations

ashleynolan:
## First beta release of v4 branch (this will merge to the v4 branch, not master, so that releases to master can still take place if needed)

Update to typography map and font declarations with fozzie.

### Changed
- **Breaking** Type map has all been updated to match new design-token names.
- Updating typography set to `JustEatBasis` and updating all typography definitions to new sizes/weights.
- Pre-loading font classes no longer needed, as the webfonts should be loaded in immediately (as part of `head` of document).

### Removed
- `Aspira` font-face declaration no longer needed.

## Typography Naming Migration Guide
- `base` > `body-s`
- `base--scaleUp` > `body-l`
- `small` > `caption`
- `mid` > `heading-s`
- `large` > `heading-m`
- `jumbo` > `heading-xl`
- `jumbo--scaleUp` > `heading-xxl`
- `large--scaleDown` > `heading-s`
- `mid--scaleUp` > No longer exists in type map. Should change to `heading-s` or `heading-m`

## UI Review Checks

- [ ] UI Documentation has been created (will be created after v4 is finalised)

## Browsers Tested

- [x] Chrome


CHANGELOG.md
```diff
@@ -4,6 +4,30 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html)
 
 
+v4.0.0-beta.0
+------------------------------
+*August 19, 2020*
```
simonsandersje, line 9: Date will need updating

src/scss/components/_media-element.scss
```diff
@@ -144,11 +144,11 @@ $mediaElement-spacing           : spacing(x2);
 
     .c-mediaElement-text {
         margin-top: 0;
-        @include font-size(small, false);
+        @include font-size(caption, false);
     }
 
     .c-mediaElement-text--large {
-        @include font-size(large, false);
+        @include font-size(heading-m, false);
```
simonsandersje, line 151: Semantically, large text isn't necessarily a heading - maybe `body-xl` would be a good addition - I know there might be some duplication, but might be a bit clearer.

src/scss/components/optional/_content-header.scss
```diff
@@ -24,13 +24,13 @@
...
             font-weight: $font-weight-base;
-            @include font-size(base--scaleUp, false);
+            @include font-size(body-l, false);
         }
 
         .c-contentHeader-subTitle {
             display: block;
             margin-top: 4px;
-            @include font-size(base, false);
+            @include font-size(body-s, false);
```
simonsandersje, line 33: Similar to above - except could this `subTitle` be `heading-s` ?

src/scss/components/optional/_content-title.scss
```diff
@@ -30,7 +30,7 @@
             margin: 0;
             width: 100%;
             font-weight: $font-weight-base;
-            @include font-size(base--scaleUp, false);
+            @include font-size(body-l, false);
```
simonsandersje, line 33: `_content-title.scss` - again `heading-s` maybe?

src/scss/objects/_lists.scss
```diff
@@ -51,12 +51,12 @@ ol {
 .c-list-ordered {
     > li {
         font-weight: $font-weight-bold;
-        @include font-size(mid);
+        @include font-size(heading-s);
```
simonsandersje, line 54: `heading-s` vs `body-l` - semantics again

simonsandersje (commented):
A few nit-picking comments

src/scss/components/_media-element.scss
DamianMullins: I agree it might be clearer to devs when styling components which font size to use if we add the extra sizes.

src/js/modules/hasFontLoaded/index.js
```diff
@@ -0,0 +1,27 @@
...
+ *
+ * @module hasFontLoaded
+ */
+
+/**
+* Init method initialises the FontFaceObserver events
+*
+*/
+
+export const hasFontLoaded = () => {
```
kevinrodrigues, line 15: Would this be better named `initLoadFont`?

src/js/modules/hasFontLoaded/test/index.test.js
```diff
@@ -0,0 +1,74 @@
...
+    describe('when the fonts do not load correctly', () => {
+        beforeAll(() => {
+            FontFaceObserver.mockImplementation(() => ({
+                load: () => new Promise((...reject) => {
+                    reject();
+                })
+            }));
+        });
+
+        it('causes FontFaceObserver to reject', () => {
```
kevinrodrigues, line 44: Should this one be testing that the `logError` is called instead?

src/js/modules/hasFontLoaded/test/index.test.js
```diff
@@ -0,0 +1,74 @@
...
+                <link href="https://d30v2pzvrfyzpo.cloudfront.net/fonts/JustEatBasis-Regular-optimised.woff2" type="font/woff2">
+                <body></body>
+            `);
+
+            // Act
+            hasFontLoaded();
+
+            // Assert
+            setTimeout(() => {
+                expect(FontFaceObserver).toBeCalled();
```
kevinrodrigues, line 29: Should use `toHaveBeenCalled` instead here as you want to check the actual mock is being called.

src/scss/components/_media-element.scss
ashleynolan: Agree – atm, this is just a lift and shift to replace the old variables with equivalent new ones from the global design tokens. As design go through each component when working on PIE, we should be able to replace these with alias variable tokens specific to each component rather than having these point at the defined heading tokens (just as that's all we have atm, they haven't looked at specific components like the `mediaElement` component yet).

I'll add a TODO comment to any of these that seem to be un-semantic so that we can make sure they get looked at (these will get ported out of fozzie as they move into the `fozzie-components` repo).

DamianMullins: Sounds good to me 👍

src/js/modules/hasFontLoaded/index.js
ashleynolan: It doesn't actually load the font – it just provides a way of observing if the font has loaded.

I'm not entirely sure if we even need this method anymore tbh, as this was mostly to monitor the loading of the font, whereas the way that we're recommending the font be included should block render via preload. I think I'll remove this and then if it's needed when we roll out, can look at it again in future.

src/scss/components/optional/_content-header.scss
ashleynolan: Have added a TODO to this to make sure this is addressed when we port the component (as it should have a different alias token at that point)

src/scss/components/optional/_content-title.scss
ashleynolan: Have added a TODO to this to make sure this is addressed when we port the component (as it should have a different alias token at that point)

src/scss/objects/_lists.scss
ashleynolan: Have added a TODO to this to make sure this is addressed when we port the component (as it should have a different alias token at that point)

src/js/modules/hasFontLoaded/index.js
ashleynolan: Have now removed this

ashleynolan:
@simonsandersje @kevinrodrigues @DamianMullins I've updated based on your feedback – let me know if there's anything else that you think needs doing.

kevinrodrigues (approved):

DamianMullins (approved):

PR #313, opened 2020-09-23
v4.1.0 - Added font-family property to headings.

DamianMullins:
### Changed
- Added `font-family` property back for headings as the font-family can be overridden by the reset styles if headings are children of elements such as buttons.


zlatinivanov-je (approved):
LGTM

benwhite-deltas (approved):

PR #315, opened 2020-10-24
v4.2.0 - Only show focus styles when needed.

DamianMullins:
### Added
- `:focus-visible` styles which provide more control over when focus outlines are displayed. This allows us to only show the focus outlines when using a keyboard. 

    > show focus styles only when they are needed, using the same heuristic that the browser uses to decide whether to show the default focus indicator

    See this article for more info - https://matthiasott.com/notes/focus-visible-is-here.


simonsandersje (approved):

kevinrodrigues (commented):
Nice 👍

kevinrodrigues (approved):

PR #324, opened 2021-03-22
v5.0.0-beta.5 - Import typography design tokens into Fozzie

1itvinka:
### Changed
- Updated type-map with new typography variables for font-size and line-height.
- Typography variables transitioned:
  - `$font-weight-base` > `$font-weight-regular`
  - `$font-weight-bold` definition was removed as it now comes from pie-design-tokens.

## UI Review Checks

- [ ] This PR has been checked with regard to our brand guidelines

### DIFF
--- a/src/scss/settings/_variables.scss
+++ b/src/scss/settings/_variables.scss
@@ -9,57 +9,55 @@
 // Set this in pixels (but do not add px),
 // the font-size mixin will convert to REMS
 
-$font-size-base             : 14;
+$font-size-base             : $font-paragraph-02;
 $font-size-base-px          : $font-size-base * 1px;
 
 // Type map – these are our global font-sizes and line-heights across the site
 $type: (
     caption: (
-        default: (12, 16)
+        default: ($font-size-a)
     ),
     body-s: (
-        default: (14, 20)
+        default: ($font-size-b)
     ),
     body-l: (
-        default: (16, 24)
+        default: ($font-size-c)
     ),
     subheading-s: (
-        default: (20, 28),
-        narrow: (16, 24)
+        default: ($font-size-d),
+        narrow: ($font-size-c)
     ),
     subheading-l: (
-        default: (24, 32),
-        narrow: (20, 28)
+        default: ($font-size-e),
+        narrow: ($font-size-d)
     ),
     heading-s: (
-        default: (20, 28),
-        narrow: (16, 24)
+        default: ($font-size-d),
+        narrow: ($font-size-c)
     ),
     heading-m: (
-        default: (24, 32),
-        narrow: (20, 28)
+        default: ($font-size-e),
+        narrow: ($font-size-d)
     ),
     heading-l: (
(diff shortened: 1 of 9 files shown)


src/scss/settings/_variables.scss
```diff
@@ -9,57 +9,55 @@
...
+        default: ($font-size-h),
+        narrow: ($font-size-g)
     )
 ) !default;
 
-$font-weight-base           : 400;
-$font-weight-bold           : 600;
-$font-weight-headings       : 600;
+$font-weight-headings       : $font-weight-bold;
 $line-height-base           : line-height();
```
1itvinka, line 57: not sure if I should also change this line

src/scss/settings/_variables.scss
```diff
@@ -9,57 +9,55 @@
...
 
 // Font stacks
-$font-family-base           : 'JustEatBasis', Arial, sans-serif;
+$font-family-base           : $font-family-primary, Arial, sans-serif;
 $font-family-mono           : Menlo, Monaco, 'Courier New', monospace;
```
1itvinka, line 61: haven't found this font family in pie design tokens. https://github.com/justeat/pie-design-tokens/blob/master/pie-design-tokens.jsonc

src/scss/settings/_variables.scss
dandel10n: line-height here uses [mixin](https://github.com/justeat/f-utils/blob/10f1daa6b77bd97a1b6083c2cfd2600f5ee2febc/src/scss/functions/_units.scss#L12) so no need to change.

src/scss/settings/_variables.scss
dandel10n: Let's leave it as it is

src/scss/settings/_variables.scss
```diff
@@ -9,57 +9,55 @@
...
-        default: (48, 56),
-        narrow: (32, 40)
+        default: ($font-size-h),
+        narrow: ($font-size-g)
     )
 ) !default;
 
-$font-weight-base           : 400;
-$font-weight-bold           : 600;
-$font-weight-headings       : 600;
```
DamianMullins, line 58: Is the plan to update these references within Fozzie in another PR? If so, do you need to remove them yet?

dandel10n: `$font-weight-bold` was removed from here as this variable now comes straight from `pie-design-tokens` so all the references can stay as they were.  And as far as I can see all the references for `$font-weight-base` were changed to use `$font-weight-regular` instead (which comes from `pie-design-tokens` as well).

1itvinka: $font-weight-bold definition was removed as it now comes from pie-design-tokens. I'll update the PR description.
$font-weight-headings is still here (line 56).

DamianMullins (approved):

dandel10n (approved):

PR #334, opened 2021-08-20
v6.0.0-beta.0 - new font

dandel10n:
### Updated
- pie-design-tokens package to v1.0.0-beta.0 to include new font vars

### Added
- `$font-family-secondary` as a fallback font (Arial)

### Changed
- `$font-weight-headings` use `$font-weight-extrabold` font weight (800)


ashleynolan:
Will review this shortly – just a note that I'll make a PR to move the base repo to v5 too when I get chance 👍

PR #335, opened 2021-08-23
v5.0.0 Release

ashleynolan:
v5.0.0
*August 23, 2021*

## Main changes:
- Colour and typography variables updated to use `pie-design-tokens` repo instead of `fozzie-colour-palette`. 
- When importing v5 fozzie, consuming applications only get included variables, functions and mixins (no classes). Component styles for buttons, form-fields etc are being deprecated as we move our components over to versionable packages, but can still be included via optional mixins applications that still need this as part of the transition.
- Two optional variables can be used to include a base set of styles (similar to v4). These are `$includeBaseFramework` and `$includeMinimalFramework`. Check out the `_templates.scss` file to see what styles are included when these are set to true.

### Changed
- Eyeglass version bumped to v3.
- Made all styles and components optional (wrapped in a mixin).
- Lots of colour variables moved over as part of using the new `pie-design-tokens` repo instead of `fozzie-colour-palette`.
- Updated type-map with new typography variables for font-size and line-height.

### Removed
- `/settings/glyphs` moved into badges, as it is the only place it is used.
- `/images` as not used.
- `kickoff-grid` dependency brought into fozzie, to remove external dependency (making it easier to change if needed).
- `$font-weight-bold` was removed as it now comes straight from `pie-design-tokens`.

** Full list of changes (including all colour variable changes) can be found in beta release CHANGELOGs (see full CHANGELOG entries) **

## UI Review Checks

- [x] UI Documentation has been updated (more incoming shortly)

### DIFF
--- a/src/scss/objects/_tables.scss
+++ b/src/scss/objects/_tables.scss
@@ -22,8 +22,8 @@
     $table-bgColor--accent        : $color-grey-10 !global; // Background color used for `.table-striped`.
     $table-border--color          : $color-grey-10 !global; // Border color for table and cell borders.
     $table-border--width          : 2px !global; // Border width for table border.
-    $table-innerBorder--color     : $color-border !global;
-    $table-verticalBorder--color  : $color-border--interactive !global;
+    $table-innerBorder--color     : $color-grey-40 !global;
+    $table-verticalBorder--color  : $color-grey-50 !global;
 
 
     /**
--- a/src/scss/objects/_tables.scss
+++ b/src/scss/objects/_tables.scss
@@ -1,97 +1,102 @@
-/**
- * Objects > Tables
- * =================================
- * Default Table styles
- *
- * File Index:
- * - Table specific variables
- * - Default table styles
- *
- * Documentation:
- * TBC
- */
+ @mixin tables() {
+    /**
+    * Objects > Tables
+    * =================================
+    * Default Table styles
+    *
+    * This is an optional component within Fozzie.
+    * If you'd like to use it in your project you can include it by adding `@include tables();` into your SCSS dependencies file.
+    *
+    * File Index:
+    * - Table specific variables
+    * - Default table styles
+    *
+    * Documentation:
+    * TBC
+    */
 
-/**
- * Define associated Table variables
- */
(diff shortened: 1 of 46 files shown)


DamianMullins (approved):
Looks good to me, and haven't noticed any issues whilst using the beta version on menu. 

One thing to mention, to future-proof us a little, could we consider removing the `$spacing` variables `x0.5` & `x1.5` as they aren't using valid syntax? Where people are using those variables they can replace it with something like `spacing() * 0.5` if they need to use the fraction values?

If not in v5 possibly in v6?

ashleynolan:
@DamianMullins Yeh – funnily enough, I think we hit this when trying to upgrade one of our dependencies in the component repo, so I think that'll happen pretty shortly 👍

kevinrodrigues (approved):

src/scss/objects/_tables.scss
```diff
@@ -1,97 +1,102 @@
...
-$table-innerBorder--color     : $color-border !default;
-$table-verticalBorder--color  : $color-border--interactive !default;
+    /**
+    * Define associated Table variables
+    */
+    $table-bgColor                : $color-container-default !global; // Default background color used for all tables.
+    $table-bgColor--accent        : $color-grey-10 !global; // Background color used for `.table-striped`.
+    $table-border--color          : $color-grey-10 !global; // Border color for table and cell borders.
+    $table-border--width          : 2px !global; // Border width for table border.
+    $table-innerBorder--color     : $color-grey-40 !global;
```
dandel10n, line 25:
```suggestion
    $table-innerBorder--color     : $color-border-strong !global;
```

src/scss/trumps/_utilities.scss
```diff
@@ -1,409 +1,411 @@
...
-.u-lightenBg {
-    background-color: $white;
-}
-
-.u-bordered,
-%u-bordered {
-    @include bordered();
-}
+    .u-lightenBg {
+        background-color: $color-white;
```
dandel10n, line 26:
```suggestion
        background-color: $color-container-default;
```

PR #338, opened 2021-09-10
v6.0.0-beta.1 - move normalize styles from base to optional

dandel10n:
### Changed
- Move normalize styles from base to optional. I had to copy the styles straight into the project to be able to make them optional wrapping them into a mixin. 

This way we will be able not to include them into every fozzie-component and only include them into consuming app if needed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,14 @@ Future Todo List
 - Make typography and utility classes silent extenders (so that they can be extended by components without importing all utility classes).
 
 
+v6.0.0-beta.1
+------------------------------
+*September 13, 2021*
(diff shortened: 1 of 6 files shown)


DamianMullins (approved):

ashleynolan (changes requested):
This'll be a breaking change for any consuming applications, so needs to be a major version bump. Might be worth including as part of the v6 changes?

dandel10n:
@ashleynolan ok, I thought that consuming apps use `$includeBaseFramework` or `$includeMinimalFramework` vars anyway. But I can add it into v6 to stay on a safe side

PR #339, opened 2021-09-13
v5.1.0 - move normalize styles from base to optional

dandel10n:
### Changed
- Move normalize styles from base to optional. We had to copy the styles straight into the project to be able to make them optional wrapping them into a mixin.
!!Note: If you don't use `$includeBaseFramework` or `$includeMinimalFramework` vars and want to have normalize styles in you project, you need to `@include normalize()` mixin starting from this version.

### DIFF
--- a/src/scss/base/_normalize.scss
+++ b/src/scss/base/_normalize.scss
@@ -1,198 +1,351 @@
 @mixin normalize() {
-    /**
-    * Normalize Styles
-    * =================================
-    * We use normalize.css for the bulk of our 'reset' (https://www.npmjs.com/package/normalize-scss).
-    * We wrap it into a mixin to be able to make it optional.
-    *
-    * This is an optional component within Fozzie.
-    * If you'd like to use it in your project you can include it by adding `@include normalize();` into your SCSS dependencies file.
-    */
+/*! normalize.css v8.0.1 | MIT License | github.com/necolas/normalize.css */
 
-    /*! normalize-scss | MIT/GPLv2 License | bit.ly/normalize-scss v7.0.1*/
-    html {
-        line-height: 1.15;
-        -ms-text-size-adjust: 100%;
-        -webkit-text-size-adjust: 100%;
-    }
-    body {
-        margin: 0;
-    }
-    article,
-    aside,
-    footer,
-    header,
-    nav,
-    section {
-        display: block;
-    }
-    h1 {
-        font-size: 2em;
-        margin: 0.67em 0;
-    }
-    figcaption,
-    figure {
-        display: block;
-    }
-    figure {
-        margin: 1em 40px;
-    }
-    hr {
-        box-sizing: content-box;
-        height: 0;
-        overflow: visible;
-    }
-    main {
-        display: block;
-    }
-    pre {
-        font-family: monospace;
-        font-size: 1em;
-    }
-    a {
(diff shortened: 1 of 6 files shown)


src/scss/base/_normalize.scss
```diff
@@ -0,0 +1,198 @@
...
+    * =================================
+    * We use normalize.css for the bulk of our 'reset' (https://www.npmjs.com/package/normalize-scss).
+    * We wrap it into a mixin to be able to make it optional.
+    *
+    * This is an optional component within Fozzie.
+    * If you'd like to use it in your project you can include it by adding `@include normalize();` into your SCSS dependencies file.
+    */
+
+    /*! normalize-scss | MIT/GPLv2 License | bit.ly/normalize-scss v7.0.1*/
+    html {
```
ashleynolan, line 13: Just a QQ – how did you generate this file? Looking at the project, I can see the 3 SCSS files, so just wondering if when new versions come out and we want to update this.

dandel10n: @ashleynolan I copied the output we have in the components.

ashleynolan: It might be worth following the NPM docs to use the approach stated on the package (can be done as a separate Engineering excellence ticket – just otherwise I'm not sure how we upgrade this if the normalise package upgrades in the future, as we won't be compiling it anywhere from the SCSS files then):

> Approach 1: Download and use normalize-scss as a starting point for your own project's base Sass, customising the values to match the design's requirements. (The best approach, IMO.)
> 
> Copy the normalize-scss files to your sass directory so that you can alter it as you include it in your project. To aid with this method, normalize-scss includes several ready-made "fork" versions: * fork-versions/default - Fork for libSass or Ruby Sass * fork-versions/typey - Fork with Typey * fork-versions/typey-chroma-kss - Fork with Typey, Chroma and KSS

DamianMullins: One other thing to consider is that the normalize-scss package appears to be using a very old version of normalize.css ([v5](https://github.com/JohnAlbin/normalize-scss#normalizecss-v5) compared to the [most recent v8](https://github.com/necolas/normalize.css)) and hasn't been updated for 4 years.

So maybe referencing `normalize-scss` is not the best approach?

DamianMullins: v8 is actually almost 3 years old itself — I'm guessing there aren't many reset updates to be made these days?

ashleynolan: Yeh, that's true. I guess we could just look to take in v8 itself then, as agree that it's not necessarily something that updates often. If we create a ticket for that, happy to merge this as-is

ashleynolan (dismissed):

DamianMullins (dismissed):

ashleynolan (approved):

src/scss/base/_normalize.scss
dandel10n: I will create a ticket for this

dandel10n: There is [a discussion](https://github.com/necolas/normalize.css/issues/856) in the normalize project about better maintained up-to-date alternatives. I will create a ticket to look into it and replace

ashleynolan (approved):

DamianMullins (approved):

PR #340, opened 2021-09-16
v6.0.0-beta.2 - Update spacing variables.

DamianMullins:
### Removed
- SCSS spacing map variables with invalid names — these names throw an error when attempting to use Fozzie with `dart-scss`. If you still need these values in consuming applications you can use the divide or multiply operator inline e.g. `margin-top: spacing() * 1.5;`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,10 +8,19 @@ Future Todo List
 - Make typography and utility classes silent extenders (so that they can be extended by components without importing all utility classes).
 
 
+v6.0.0-beta.2
+------------------------------
+*September 17, 2021*
+
+### Removed
+- SCSS spacing map variables with invalid names — these names throw an error when attempting to use Fozzie with `dart-scss`. If you still need these values in consuming applications you can use the divide or multiply operator inline e.g. `margin-top: spacing() * 1.5;`
+
+
 v6.0.0-beta.1
 ------------------------------
 *September 15, 2021*
 
+### Changed
 - Bump the version to include changes from v5.1.0 to beta version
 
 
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -38,7 +47,7 @@ v5.1.0
 
 ### Changed
 - Move normalize styles from base to optional. We had to copy the styles straight into the project to be able to make them optional wrapping them into a mixin.
-!!Note: If you don't use `$includeBaseFramework` or `$includeMinimalFramework` vars and want to have normalize styles in you project, you need to `@include normalize()` mixin starting from this version. 
+!!Note: If you don't use `$includeBaseFramework` or `$includeMinimalFramework` vars and want to have normalize styles in you project, you need to `@include normalize()` mixin starting from this version.
```
zlatinivanov-je, line 50: 👍  I like that clarification, here.

zlatinivanov-je (approved):

src/scss/settings/_variables.scss
```diff
@@ -104,9 +104,7 @@ $spacing: 8px; // only for use in map below – not for in components
 // spacing map
 // Can access levels by using the spacing function > spacing() or spacing(x5) for example
 $spacing: (
-    x0.5 : $spacing / 2,
```
ashleynolan, line 107: Out of interest, what's the character that makes this invalid?

I think when we come to add the design tokens for spacing from the PIE design tokens, this issue will disappear as they'll be set there instead.

Could we possibly hold off on this change until after the colour updates (we'll be putting them in this week/early next week). Then we could potentially just update the design tokens to have the spacing values and replace this whole spacing map entirely? ([here are the tokens that it'll be changing too](https://docs.google.com/spreadsheets/d/1KRVJgMELOA-QkKuqNZh-9z4kpQe__diuRhjIRLxNR1Q/edit#gid=1188492957)).

ashleynolan (changes requested):
Just question about if we csan hold this until next week potentially.

src/scss/settings/_variables.scss
DamianMullins: It's the dot in the key name that is invalid.

DamianMullins: Sure, can hold off. I was just looking at the issues in using Vite to compile our Vue packages again and this is the first thing that is flagged up when using Vite with dart-scss.

ashleynolan: Makes sense. I wonder if it would make it valid to change it to be `'x0.5' : $spacing / 2,` potentially?

May fix the parsing issue then if that's valid syntax.

ashleynolan: Ah, ok. That makes sense. I'll try and have a look at this and replace these spacing variables with the design tokens, as they should have new names which will solve this issue (and save us having to make 2 similar breaking changes then).

PR #343, opened 2021-09-23
v6.0.0-beta.4 - new button styles for icing phase 2

dandel10n:
### Changed
- Button styles in line with icing phase 2.
Now button modifiers are:
- Btn Background Colour modifiers: `o-btn--primary`, `o-btn--secondary`, `o-btn--outline`, `o-btn--ghost`.
- Btn Size Modifiers: `o-btn--sizeLarge`, `o-btn--sizeSmall`, `o-btn--sizeXSmall`.
- Btn Layout Modifiers: `o-btn--block`, `o-btn--wide`, `o-btn--icon`, `o-btn--fullWidth`, `o-btn--disabled`.
- Moved font-family declaration from the button component styles to typography styles.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@ Future Todo List
 
 v6.0.0-beta.4
 ------------------------------
-*September 23, 2021*
+*September 27, 2021*
 
 ### Added
 - `o-btn--outline`, `o-btn--ghost`, `o-btn--fullWidth` modifiers.
@@ -19,7 +19,7 @@ v6.0.0-beta.4
 ### Removed
 - `o-btn--tertiary`. Use `o-btn--outline` instead.
 - `o-btn--block`. Use `o-btn--fullWidth` instead.
-- `o-btn--wide`. 
+- `o-btn--wide`.
 
 ### Changed
 - Button styles in line with icing phase 2.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,13 +18,16 @@ v6.0.0-beta.4
 
 ### Removed
 - `o-btn--tertiary`. Use `o-btn--outline` instead.
+- `o-btn--block`. Use `o-btn--fullWidth` instead.
+- `o-btn--wide`. 
 
 ### Changed
 - Button styles in line with icing phase 2.
+- renamed `o-btnLink` to `o-btn--link`.
 Now button modifiers are:
 - Btn Background Colour modifiers: `o-btn--primary`, `o-btn--secondary`, `o-btn--outline`, `o-btn--ghost`.
 - Btn Size Modifiers: `o-btn--sizeLarge`, `o-btn--sizeSmall`, `o-btn--sizeXSmall`.
-- Btn Layout Modifiers: `o-btn--block`, `o-btn--wide`, `o-btn--icon`, `o-btn--fullWidth`, `o-btn--disabled`.
+- Btn Layout Modifiers: `o-btn--icon`, `o-btn--fullWidth`, `o-btn--disabled`, `o-btn--link`.
 - Moved font-family declaration from the button component styles to typography styles.
 
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,19 @@ Future Todo List
 - Make typography and utility classes silent extenders (so that they can be extended by components without importing all utility classes).
 
 
+v6.0.0-beta.4
+------------------------------
+*September 23, 2021*
+
+### Changed
+- Button styles in line with icing phase 2.
+Now button modifiers are:
+- Btn Background Colour modifiers: `o-btn--primary`, `o-btn--secondary`, `o-btn--outline`, `o-btn--ghost`.
+- Btn Size Modifiers: `o-btn--sizeLarge`, `o-btn--sizeSmall`, `o-btn--sizeXSmall`.
+- Btn Layout Modifiers: `o-btn--block`, `o-btn--wide`, `o-btn--icon`, `o-btn--fullWidth`, `o-btn--disabled`.
+- Moved font-family declaration from the button component styles to typography styles.
+
+
 v6.0.0-beta.3
 ------------------------------
 *September 22, 2021*
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,13 @@ v6.0.0-beta.4
 ------------------------------
 *September 23, 2021*
 
+### Added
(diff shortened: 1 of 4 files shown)


src/scss/objects/_buttons.scss
```diff
@@ -54,48 +64,52 @@
     * 3. Make buttons inherit font styles.
     * 4. Force all elements using beautons to appear clickable.
     * 5. Normalise box model styles.
-    * 6. If the button’s text is 1em, and the button is (3 * font-size) tall, then
-    *    there is 1em of space above and below that text. We therefore apply 1em
-    *    of space to the left and right, as padding, to keep consistent spacing.
-    * 7. Fixes odd inner spacing in IE7.
+    * 6. Fixes odd inner spacing in IE7.
```
ashleynolan, line 67: We can probably delete this line, if it's for an IE7 fix?

src/scss/base/_typography.scss
```diff
@@ -179,6 +179,13 @@
         }
     }
 
+    /**
+    * Buttons
+    */
+    button {
+        font-family: $font-family-base;
```
ashleynolan, line 186: Just out of interest – do we need this to override the browser default as it doesn't inherit from the font family set on the main document?

CHANGELOG.md
```diff
@@ -8,6 +8,19 @@ Future Todo List
...
 
 
+v6.0.0-beta.4
+------------------------------
+*September 23, 2021*
+
+### Changed
+- Button styles in line with icing phase 2.
+Now button modifiers are:
+- Btn Background Colour modifiers: `o-btn--primary`, `o-btn--secondary`, `o-btn--outline`, `o-btn--ghost`.
```
ashleynolan, line 18: It would be useful to list the specific things that have been removed so that anyone updating knows specifically about these breaking changes.

The main one I can see is `o-btn--tertiary` being removed, so it's worth calling that out under a removed sections here.

Similarly, I'd put the button size modifiers, `o-btn--outline` and `o-btn--ghost` under an `Added` section, just so it's clear what got added and removed in each release. This can be in addition to the Changed block you've already added, just makes it super clear what is new, and what is bein removed.

src/scss/base/_typography.scss
dandel10n: I tested it and when the font for button is not specified it defaults back to browser default even with body font being ours. I didn't spent much time investigating and just added this style in to clear font from `o-btn` element and to still stay safe.

src/scss/objects/_buttons.scss
dandel10n: Yeap, tested and removed

src/scss/base/_typography.scss
DamianMullins: I hit this recently in menu when working on the phase 1 icing changes where all buttons that didn't have the type-face set explicitly would revert back to use the browser default font (e.g. `sans` in Chrome).

src/scss/objects/_buttons.scss
```diff
@@ -10,92 +10,101 @@
...
         user-select: none;
-
-        color: $color-grey-50;
+        color: $btn-default-color;
         text-decoration: none;
 
-        &:hover,
-        &:focus {
-            background-color: $btn-default-bgColor--hover;
+        // Hide focus styles if they're not needed, for example, when an element receives focus via the mouse.
```
zlatinivanov-je, line 84: Could that specificity be simplified if declared after `&:focus-visible` e.g. `&:focus`

zlatinivanov-je (commented):
Happy to approve, subject to comments. 
Also should there be a `package.json` version update?

src/scss/objects/_buttons.scss
```diff
@@ -10,92 +10,101 @@
...
+        &:focus:not(:focus-visible) {
```
zlatinivanov-je, line 85: Could that specificity be simplified if declared after `&:focus-visible` e.g. `&:focus`

DamianMullins: The styles are written in this way to progressively enhance browsers that support `:focus-visible`. There is some [more info in this article](https://matthiasott.com/notes/focus-visible-is-here) and [a PR where this has been added in the past here](https://github.com/justeat/fozzie/pull/315).

The styles in that layout file linked above would apply to all elements (unless explicitly overridden) which means that if users don't pull those styles in there could be inconsistent visual outlines on the page though i.e. buttons would only have outlines when the user is using a keyboard but other elements would always have outline.

dandel10n: I removed theses styles for now for consistency. If consuming app uses layout styles from fozzie then buttons will inherit them, if not it will be up to consuming app to add those styles for all the interactive elements or not

src/scss/objects/_buttons.scss
```diff
@@ -268,23 +323,37 @@
...
 
     /**
     * Disabled button
     */
 
-    .o-btn--disabled {
-        cursor: default;
+    .o-btn--disabled,
+    .o-btn.disabled,
+    &.is-disabled,
```
DamianMullins, line 344: Can you use `&` here? I don't think we're nested inside another style unless I'm missing something?

DamianMullins: From my testing it's applying the style to any element that has the `.is-disabled` class or the `disabled` attribute.

dandel10n: Good catch, thanks, fixed!

zlatinivanov-je (approved):

DamianMullins (approved):

src/scss/objects/_buttons.scss
```diff
@@ -10,92 +10,92 @@
...
         user-select: none;
-
-        color: $color-grey-50;
+        color: $btn-default-color;
         text-decoration: none;
 
-        &:hover,
         &:focus {
-            background-color: $btn-default-bgColor--hover;
+            border: 2px solid $btn-default-outline-color;
```
DamianMullins, line 83: I think this should be `outline` rather than `border` as the border affects the position of the button?

src/scss/objects/_buttons.scss
```diff
@@ -10,92 +10,92 @@
...
         user-select: none;
-
-        color: $color-grey-50;
+        color: $btn-default-color;
         text-decoration: none;
 
-        &:hover,
         &:focus {
-            background-color: $btn-default-bgColor--hover;
+            box-shadow: 0 0 0 2px solid $btn-default-outline-color;
```
DamianMullins, line 83: This isn't a valid value for `box-shadow`, the "solid" part needs to be removed.

PR #346, opened 2021-10-04
v6.0.0-beta.7 - Component style updates.

DamianMullins:
### Added
- `c-card--info` component modifier.
- `c-rating-description` component modifier.

### Changed
- Set default paragraph size to match base font-size.
- Card border colour matches PIE designs.
- `c-mediaElement--fullstack--negativeTop` modifier top margin updated.
- Badge component font-size explicitly set.
- `c-badge--indicator` modifier colours updated.

### Fixed
- Breadcrumb icon size.

### DIFF
--- a/src/scss/components/_media-element.scss
+++ b/src/scss/components/_media-element.scss
@@ -134,7 +134,7 @@
 
             .c-mediaElement--fullstack--negativeTop & {
                 flex: auto;
-                margin-top: -#{spacing(x2)};
+                margin-top: spacing(x4);
             }
 
             // some content uses br for content shaping – get rid of it on mobile devices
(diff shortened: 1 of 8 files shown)


src/scss/components/_media-element.scss
```diff
@@ -134,7 +134,7 @@
 
             .c-mediaElement--fullstack--negativeTop & {
                 flex: auto;
-                margin-top: -#{spacing(x2)};
+                margin-top: spacing(x4);
```
dandel10n, line 137: The modifier part of the class name states `--negativeTop` and with this change it is not anymore. Should we rename this class name or add a new one with `spacing(x4)` top margin value instead?

dandel10n (approved):

simonsandersje (approved):

PR #358, opened 2021-12-20
v6.1.0 - remove theme var

dandel10n:
### Removed
- `$theme` variable (`_theme.scss` file) as there are no more theme overrides (like typography or specific styles for ml).


AlixRichards (approved):

ashleynolan:
I think this change will also need the legacy `f-searchbox` to be updated, as there's still a reference to [this theme variable in there](https://github.com/justeat/f-searchbox/blob/7dd7503c0762ad815d6f8cb1108f109bb42b7558/src/assets/scss/common.scss#L3-L8).

ashleynolan:
@dandel10n One other place that will also need updating – [Legacy f-header](https://github.com/justeat/f-header/blob/f662ef536b1dc25ce698c02d83957ab0a4d28509/src/scss/partials/_logo.scss#L18-L31)

PR #359, opened 2022-01-25
V7.0.0 update tokens

kevinrodrigues:
### Changed
- **Breaking** Type map for $spacing to reference pie design tokens.
  - If you are using the below mapping values please note they have changed or been removed:
    $spacing x 6 = spacing(h) instead of spacing(x6) - Note the size change has gone from 48px > 56px.
    $spacing x 9 - Has been removed - Please map to the next value up or down e.g spacing(j) or spacing(i).

| Old map usage | New map usage |
| ----- | ----- |
| spacing(x0.5) | $spacing(a) |
| spacing(base) | $spacing(b) |
| spacing(x1.5) | $spacing(c) |
| spacing(x2) | $spacing(d) |
| spacing(x3) | $spacing(e) |
| spacing(x4) | $spacing(f) |
| spacing(x5) | $spacing(g) |
| spacing(x6) | $spacing(h) |
| spacing(x7) | $spacing(h) |
| spacing(x8) | $spacing(i) |
| spacing(x9) | Removed, replace with closet size up or down |
| spacing(x10) | $spacing(j) |

- `pie-design-tokens` package.json version to bring in new elevation changes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,16 @@ Future Todo List
 ------------------------------
 - Make typography and utility classes silent extenders (so that they can be extended by components without importing all utility classes).
 
+v7.0.0
+------------------------------
+*January 25, 2022*
+
+### Changed
+- **Breaking** Type map for $spacing to reference pie design tokens.
+  - If you are using the below mapping values please note they have changed or been removed:
+    $spacing x 6 = spacing(h) instead of spacing(x6) - Note the size change has gone from 48px > 56px in the new mapping.
+    $spacing x 9 - Has been removed.
+
 
 v6.2.0
 ------------------------------
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,22 @@ v7.0.0
 - **Breaking** Type map for $spacing to reference pie design tokens.
   - If you are using the below mapping values please note they have changed or been removed:
     $spacing x 6 = spacing(h) instead of spacing(x6) - Note the size change has gone from 48px > 56px.
-    $spacing x 9 - Has been removed.
+    $spacing x 9 - Has been removed - Please map to the next value up or down e.g spacing(j) or spacing(i).
+
+| Old map usage | New map usage |
+| ----- | ----- |
+| spacing(x0.5) | $spacing(a) |
+| spacing(base) | $spacing(b) |
+| spacing(x1.5) | $spacing(c) |
+| spacing(x2) | $spacing(d) |
+| spacing(x3) | $spacing(e) |
+| spacing(x4) | $spacing(f) |
+| spacing(x5) | $spacing(g) |
+| spacing(x6) | $spacing(h) |
+| spacing(x7) | $spacing(h) |
+| spacing(x8) | $spacing(i) |
+| spacing(x9) | Removed, replace with closet size up or down |
+| spacing(x10) | $spacing(j) |
 
 - `pie-design-tokens` package.json version to bring in new elevation changes.
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,9 +14,10 @@ v7.0.0
 ### Changed
 - **Breaking** Type map for $spacing to reference pie design tokens.
   - If you are using the below mapping values please note they have changed or been removed:
-    $spacing x 6 = spacing(h) instead of spacing(x6) - Note the size change has gone from 48px > 56px in the new mapping.
+    $spacing x 6 = spacing(h) instead of spacing(x6) - Note the size change has gone from 48px > 56px.
     $spacing x 9 - Has been removed.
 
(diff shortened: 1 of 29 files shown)


CHANGELOG.md
```diff
@@ -7,6 +7,17 @@ Future Todo List
 ------------------------------
 - Make typography and utility classes silent extenders (so that they can be extended by components without importing all utility classes).
 
+v7.0.0
+------------------------------
+*January 25, 2022*
+
+### Changed
```
ashleynolan, line 14: Can you add the new mappings here too - so x2 now equals b (or whatever it actually is). Will just make the upgrade easier for when people need to replace those values in their repos

kevinrodrigues: Yup will do.

src/scss/settings/_variables.scss
```diff
@@ -97,28 +97,22 @@ $responsive-gutters           : true !default; // Disable this if you don't want
...
-    x2   : $spacing * 2,
-    x3   : $spacing * 3,
-    x4   : $spacing * 4,
-    x5   : $spacing * 5,
-    x6   : $spacing * 6,
-    x7   : $spacing * 7,
-    x8   : $spacing * 8,
-    x9   : $spacing * 9,
-    x10   : $spacing * 10
+    a    : $spacing-a,
```
ashleynolan, line 103: I know it'll be potentially flakey to do this, but would it be handy to potentially comment here what the values are (as I would think they'll rarely change in the future). Just thinking it may save people some time looking up the variables for these values – what do you think?

```suggestion
    a    : $spacing-a, // 4px
```

src/scss/settings/_variables.scss
```diff
@@ -97,28 +97,22 @@ $responsive-gutters           : true !default; // Disable this if you don't want
...
+    e    : $spacing-e,
+    f    : $spacing-f,
+    g    : $spacing-g,
+    h    : $spacing-h,
+    i    : $spacing-i,
+    j    : $spacing-j
 );
 
-$baseline: spacing(x2); // default spacing between blocks
+$baseline: spacing(d); // default spacing between blocks
```
ashleynolan, line 115: I think we should look to remove `$baseline` and replace it with the equivalent `$spacing(d)` value now where it's used in fozzie

src/scss/settings/_variables.scss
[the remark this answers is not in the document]
ashleynolan: The main one is that we still use `spacing()` for 8px (or `spacing(b)`) as a shortcut for what seems to be the most common spacing usage. I'm not tied to keeping it if we think it's clearer without it though.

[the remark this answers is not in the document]
ashleynolan: The other thing I just thought about that could come in useful is that we're likely going to switch to CSS variables shortly. If we do that, then I would think using the function could mean we don't then have to change all the sass references to the new values?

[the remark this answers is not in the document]
DamianMullins: Thanks, makes sense. I'm happy to keep it as is.

kevinrodrigues: Makes sense, ill add that in.

CHANGELOG.md
kevinrodrigues: Added a table showing old vs new.

src/scss/settings/_variables.scss
kevinrodrigues: Fixed.

src/scss/settings/_variables.scss
kevinrodrigues: Fixed.

DamianMullins (approved):

src/scss/base/_typography.scss
```diff
@@ -78,8 +78,8 @@
     h1,
     .alpha {
         @include font-size(heading-xl, true, narrow);
-        margin-top: $baseline * 2;
-        margin-bottom: $baseline;
+        margin-top: spacing(d) * 2;
```
ashleynolan, line 81: Could use `spacing(f)` here and get rid of the `* 2`

src/scss/base/_typography.scss
```diff
@@ -120,14 +120,14 @@
     // Only give these headings a margin-top if they are after other elements
     * + h2,
     * + .beta {
-        margin-top: $baseline * 2;
+        margin-top: spacing(d) * 2;
```
ashleynolan, line 123: Same here as above – just use `spacing(f)`?

src/scss/base/_typography.scss
kevinrodrigues: Fixed.

src/scss/base/_typography.scss
kevinrodrigues: Fixed.

DamianMullins (approved):

ashleynolan (approved):

PR #367, opened 2022-05-12
v8.1.0 - port over f-utils scss.

kevinrodrigues:
### Added
- f-utils `functions`, `helpers` & `mixins` in order to depreciate `f-utils` in favour of `fozzie`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,7 +2,7 @@
   "name": "@justeat/fozzie",
   "title": "Fozzie – Just Eat UI Web Framework",
   "description": "UI Web Framework for the Just Eat Global Platform",
-  "version": "8.0.0",
+  "version": "8.1.0",
   "main": "dist/js/index.js",
   "files": [
     "dist/js",
@@ -33,7 +33,6 @@
     "node": ">=10.0.0"
   },
   "dependencies": {
-    "@justeat/f-utils": "3.5.0",
     "@justeat/pie-design-tokens": "1.4.0",
     "include-media": "1.4.10"
   },
--- a/package.json
+++ b/package.json
@@ -30,7 +30,7 @@
   },
   "license": "Apache-2.0",
   "engines": {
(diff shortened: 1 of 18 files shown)


package.json
```diff
@@ -33,7 +33,6 @@
     "node": ">=10.0.0"
```
ashleynolan, line 33: Any chance we can update this node version as well? Looks a bit out of date

src/scss/tools/_index.scss
```diff
@@ -1,12 +1,12 @@
 // =================================
-// Tools
+// Fozzie SCSS Utilities
```
ashleynolan, line 2: I'd keep this as tools, as it's the index of that folder

```suggestion
// Tools
```

src/scss/_dependencies.scss
```diff
@@ -26,7 +26,7 @@
 @import '@justeat/pie-design-tokens/dist/jet';
 
 @import 'settings/variables';
-@import '@justeat/f-utils/src/scss/helpers/breakpoints'; // breakpoint helper – https://github.com/justeat/f-utils/blob/96508ae19809c4b8b83cebb55b9d74728d01c1a5/src/scss/helpers/_breakpoints.scss
+@import 'tools/helpers/breakpoints'; // breakpoint helper
```
ashleynolan, line 29: Would be good to remove this by default and add it into the `_templates` as I think we only really use this on some of our older applications.

Would make this change a breaking change then though, so maybe could you add a `TODO` to this line that we'll remove/move it as part of the next major version bump?

ashleynolan (commented):
Looks good, just a couple of comments

kevinrodrigues:
> Looks good, just a couple of comments

Thanks. Have added the suggested changes now 👍

ashleynolan (approved):

FayeCarter (approved):

PR #371, opened 2022-06-15
v9.0.0-beta.2 - update @use rules in _type.scss

jamieomaguire:
v9.0.0-beta.2
*June 15, 2022*

### Changed
- Use a `@use` rule for `line-height` function in `_type.scss`
- Remove namespace aliases from `@use` rules in `_type.scss`

### DIFF
--- a/src/scss/tools/mixins/_type.scss
+++ b/src/scss/tools/mixins/_type.scss
@@ -1,6 +1,6 @@
 @use '../../settings/variables' as v;
-@use '../functions/strip-units' as functions;
-@use '../functions/units' as units;
+@use '../functions/strip-units' as *;
+@use '../functions/units' as *;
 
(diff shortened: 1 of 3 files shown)


src/scss/tools/mixins/_type.scss
```diff
@@ -1,5 +1,6 @@
 @use '../../settings/variables' as v;
 @use '../functions/strip-units' as functions;
```
kevinrodrigues, line 2: Could you change this to `@use '../functions/strip-units'` and then change the references below please, I must have missed this one. The the next line can just be `@use '../functions/units'`

jamieomaguire: Good spot - will get on this shortly!

DamianMullins (approved):

PR #376, opened 2022-07-01
fozzie@v9.0.0-beta.6 - Scss unit testing

jamieomaguire:
v9.0.0-beta.6
*July 01, 2022*

### Added
- SCSS unit testing capabilities using `sass-true`
- New unit testing directory (to avoid packaging tests with npm)
### Changed
- yarn test command specifies which `jest.config.js` file to use


ashleynolan (commented):
Looks good. Some really minor comments, otherwise LGTM.

src/test/scss/tools/functions/px-to-em.spec.scss
```diff
@@ -0,0 +1,91 @@
+@use 'true' as *;
+@use '../../../../scss/tools/functions/px-to-em' as systemUnderTest;
```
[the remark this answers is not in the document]
jamieomaguire: Oh nice, it does work! Will commit shortly :)

.eslintrc.js
```diff
@@ -1,6 +1,7 @@
 module.exports = {
     extends: '@justeat/eslint-config-fozzie',
     rules: {
+        'import/no-extraneous-dependencies': 'off',
```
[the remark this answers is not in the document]
jamieomaguire: I can't seem to reproduce the eslint issue from before so I will remove this override for now and add back as a comment if we see any further issues

src/test/scss/scss.spec.js
```diff
@@ -0,0 +1,40 @@
+/**
+ * This is a test setup file that enables Jest to run Scss unit tests using the sass-true library
```
xander-marjoram, line 2: Nit: As this is a setup file, would it make more sense to name it `scss.setup.js`, or is this the only way to get it to run?

src/test/scss/tools/mixins/type.spec.scss
```diff
@@ -0,0 +1,39 @@
...
+        $sizeValue: 10.5;
+
+        // act & assert
+        @include assert {
+            @include output {
+                @include systemUnderTest.rem($property, $sizeValue);
+            }
+
+            @include expect {
+                padding: 11px;
```
xander-marjoram, line 17: Is it expected to round 10.5 to 11? Or is that what you're testing? (If so, might be worth adding it to the test description)

src/test/scss/scss.spec.js
jamieomaguire: I might be able to get that to work but it would require tweaking Jest to look for a file matching that pattern. Personally I'd prefer to keep the `.spec` but what if I renamed it to something like `scss-setup.spec.js`?

src/test/scss/tools/mixins/type.spec.scss
jamieomaguire: Yeah that's it - I'll rename the test

jamieomaguire: Yeah that's part of what's tested - will update the test description!

src/test/scss/tools/mixins/type.spec.scss
```diff
@@ -0,0 +1,39 @@
...
+                padding: 0.65625rem
+            }
+        }
+    }
+
+    @include it('outputs the expected CSS by rounding the default `$sizeValue` when no `$sizeValue` argument is provided') {
```
xander-marjoram, line 23: This one isn't rounding, right?
```suggestion
    @include it('outputs the expected CSS using the default `$sizeValue` when no `$sizeValue` argument is provided') {
```

PR #393, opened 2022-08-25
v9.3.0 - fixes spacing & zIndex references

kevinrodrigues:
### Fixed
- `spacing` & `zIndex` references.

### DIFF
--- a/src/scss/base/_layout.scss
+++ b/src/scss/base/_layout.scss
@@ -2,6 +2,7 @@
 @use '../settings/variables' as v;
 @use '../settings/include-media' as *;
 @use '../tools/mixins/type';
+@use '../tools/functions/spacing' as spacing;
 
 @mixin layout() {
     /**
@@ -68,14 +69,14 @@
     }
 
     .l-innerContainer--verticalSpacing {
-        margin: spacing(f) 0;
+        margin: spacing.spacing(f) 0;
 
         @include media('>=mid') {
-            margin: spacing(g) auto spacing(h);
+            margin: spacing.spacing(g) auto spacing.spacing(h);
         }
 
         @include media('>=wide') {
-            margin: spacing(i) auto;
+            margin: spacing.spacing(i) auto;
         }
     }
 
@@ -83,18 +84,18 @@
     * Layout class for containers that contain content (such as T&cs and Privacy pages)
     */
     .l-content {
-        padding-top: spacing(d);
+        padding-top: spacing.spacing(d);
 
         @include media('>mid') {
-            padding-top: spacing(f);
+            padding-top: spacing.spacing(f);
         }
     }
 
         .l-content-header {
-            margin: 0 0 spacing(d);
+            margin: 0 0 spacing.spacing(d);
 
             @include media('>mid') {
-                margin: spacing(d) 0 spacing(f);
+                margin: spacing.spacing(d) 0 spacing.spacing(f);
             }
 
             > h1 {
--- a/src/scss/base/_typography.scss
+++ b/src/scss/base/_typography.scss
@@ -1,6 +1,7 @@
 @use '../settings/variables' as v;
 @use '@justeat/pie-design-tokens/dist/jet' as dt;
 @use '../tools/mixins/type';
+@use '../tools/functions/spacing' as spacing;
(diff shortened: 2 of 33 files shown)


jamieomaguire (dismissed):

src/scss/base/_layout.scss
```diff
@@ -2,6 +2,7 @@
 @use '../settings/variables' as v;
 @use '../settings/include-media' as *;
 @use '../tools/mixins/type';
+@use '../tools/functions/spacing' as spacing;
```
DamianMullins, line 5: Minor thing, you can leave the `as spacing` part off and it'll use `spacing` as the module name, like does `type` above.

```suggestion
@use '../tools/functions/spacing';
```

src/scss/components/optional/_cookie-warning.scss
```diff
@@ -1,5 +1,7 @@
 @use '@justeat/pie-design-tokens/dist/jet' as dt;
 @use '../../tools/mixins/type';
+@use '../../tools/functions/spacing' as spacing;
+@use '../../tools/functions/zIndex' as zIndex;
```
DamianMullins, line 4: Same as my last comment, you don't need to specify the `as` part.

DamianMullins (commented):
I just checked this branch locally and it looks like there are still some modules which need updating.

 * `_alerts.scss`
 * `_listings-skeleton.scss`
 * `_order-card.scss`
 * `_buttons.scss`
 * `_form-toggle.scss`

DamianMullins (approved):

src/scss/base/_layout.scss
jamieomaguire, line 5: can probably remove the 'as' from this too :)

src/scss/base/_typography.scss
```diff
@@ -1,6 +1,7 @@
 @use '../settings/variables' as v;
 @use '@justeat/pie-design-tokens/dist/jet' as dt;
 @use '../tools/mixins/type';
+@use '../tools/functions/spacing' as spacing;
```
jamieomaguire, line 4: and here

src/scss/components/_alerts.scss
```diff
@@ -1,5 +1,6 @@
 @use '@justeat/pie-design-tokens/dist/jet' as dt;
 @use '../tools/mixins/alerts';
+@use '../tools/functions/spacing' as spacing;
```
jamieomaguire, line 3: and here

src/scss/components/_breadcrumbs.scss
```diff
@@ -1,6 +1,7 @@
 @use '@justeat/pie-design-tokens/dist/jet' as dt;
 @use '../settings/include-media' as *;
 @use '../tools/mixins/type';
+@use '../tools/functions/spacing' as spacing;
```
jamieomaguire, line 4: and here

src/scss/components/_cards.scss
```diff
@@ -1,6 +1,7 @@
 @use '@justeat/pie-design-tokens/dist/jet' as dt;
 @use '../settings/include-media' as *;
 @use '../tools/mixins/truncate';
+@use '../tools/functions/spacing' as spacing;
```
jamieomaguire, line 4: and here

src/scss/base/_layout.scss
kevinrodrigues: Will be creating a ticket to tidy this all up. Once we've agreed on a convention.

jamieomaguire (approved):

src/scss/base/_layout.scss
ashleynolan: I'm not 100% sure if it'll work, but would it be possible to potentially do:

```scss
@use '../tools/functions/index as functions;
```

and then reference `functions.spacing()`

I think it should work with that syntax and make the namespaces a bit cleaner?

kevinrodrigues: I can give this a try.

kevinrodrigues: This is now fixed here https://github.com/justeat/fozzie/pull/401. I took the function route it worked well 👍

PR #396, opened 2022-09-01
v9.3.3 - catch SCSS errors in test

jamieomaguire:
v9.3.3
*September 1, 2022*

### Added
- A basic integration test to catch SCSS errors

### DIFF
--- a/src/test/scss/fozzie.spec.scss
+++ b/src/test/scss/fozzie.spec.scss
@@ -1,0 +1,139 @@
+@use 'true' as *;
+@use 'src/scss/fozzie' as systemUnderTest;
+
+@include describe('Fozzie') {
+    @include it('does not throw any errors') {
+        // base
+        @include systemUnderTest.grid();
+        @include systemUnderTest.layout();
+        @include systemUnderTest.normalize();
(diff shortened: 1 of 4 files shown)


src/test/scss/fozzie.spec.scss
```diff
@@ -0,0 +1,139 @@
+@use 'true' as *;
+@use 'src/scss/fozzie' as systemUnderTest;
+
+@include describe('Fozzie') {
+    @include it('does not throw any errors') {
```
ashleynolan, line 5: Would there be any additional value in potentially splitting out this test into subsections (multiple it calls).

May help when there is an error thrown just to put these into logical test sections

kevinrodrigues: I think that would be good. Might make the file a bit larger but the error will help pinpoint the thing that is failing plus would be easier to debug with X

jamieomaguire: Yeah happy to - will give that a go now!

kevinrodrigues (approved):

siggerzz (approved):

PR #397, opened 2022-09-01
v10.0.0 - upgrade design tokens to version 2.0.0

jamieomaguire:
v10.0.0
*September 1, 2022*

### Changed
- **Breaking** Upgraded `@justeat/pie-design-tokens` to `v2.0.0`
- **Breaking** Replaced `$color-grey` variants with new aliases

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,7 @@ v10.0.0
 *September 1, 2022*
 
 ### Changed
-- **Breaking** Upgraded `@justeat/pie-design-tokens` to `v2.0.0`
+- **Breaking** Upgraded `@justeat/pie-design-tokens` to `v2.0.0`. [Link to changelog](https://github.com/justeat/pie-design-tokens/blob/master/CHANGELOG.md#v200)
 - **Breaking** Replaced `$color-grey` variants with new aliases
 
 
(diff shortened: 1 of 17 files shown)


CHANGELOG.md
```diff
@@ -9,6 +9,15 @@ Future Todo List
 - Update to use latest v2 PIE design tokens
 
 
+v10.0.0
```
kevinrodrigues, line 12: More of a question but most of the changes are reference changes (no impact to how this package is used in consuming apps) should we therefore us a minor bump? @ashleynolan.

ashleynolan: If consumers of fozzie don't need to change anything, then it should just be a minor version bump.

I suspect that as users rely on `pie-design-tokens` though and that has breaking changes, we'll likely need to bump the major version for those changes (if someone is referencing `$color-grey` for instance, they'll need to update stuff on their side.

CHANGELOG.md
```diff
@@ -9,6 +9,15 @@ Future Todo List
...
+------------------------------
+*September 1, 2022*
+
+### Changed
+- **Breaking** Upgraded `@justeat/pie-design-tokens` to `v2.0.0`
```
ashleynolan, line 17: Worth linking to the pie-design-tokens changelog here, as that'll tell people what's changed that they need to be aware of.

ashleynolan (dismissed):

kevinrodrigues (dismissed):

PR #398, opened 2022-09-15
v10.1.0 - Adds rating styles back in.

kevinrodrigues:
### Added
- Ratings `extend` back in as a short term fix for consuming applications.

### DIFF
--- a/src/scss/components/optional/_ratings.scss
+++ b/src/scss/components/optional/_ratings.scss
@@ -21,12 +21,8 @@
     $star-size--medium          : 28px;
     $star-size--large           : 42px;
 
-  // https://github.com/justeat/f-icons/blob/master/icons/star-filled.svg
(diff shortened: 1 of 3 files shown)


xander-marjoram (approved):

DamianMullins (approved):

src/scss/components/optional/_ratings.scss
```diff
@@ -21,12 +21,8 @@
     $star-size--medium          : 28px;
     $star-size--large           : 42px;
 
-  // https://github.com/justeat/f-icons/blob/master/icons/star-filled.svg
-  @warn 'To use the rating component you will need to add references to the SVGs in f-icons';
```
ashleynolan, line 25: Just a note – the consuming application will still need to add the silent extend mixin definition for this to work (like Menu does), so it may be worth updating the comment for that?

PR #403, opened 2022-10-05
v10.6.0 - CSS compilation and snapshot testing

jamieomaguire:
v10.6.0
*October 6, 2022*

### Added
- SCSS compilation utilities
- Snapshot test of compiled Fozzie CSS
- A (disabled) validity test of compiled Fozzie CSS using StyleLint (to be enabled in a future piece of work)
- Disable `eslint` extraneous-dependencies rule for project

### Changed
- Fixed an `em()` call in `_modal.scss` to be called with the `functions` namespace
- Fixed a number of `line-height()` calls in `_order-card.scss` to be called with the `functions` namespace
- Removed an invalid `solid` used as part of a `box-shadow` declaration in `_buttons.scss` (invalid CSS)
- Changed declaration orders of `$line-height-base` and `line-height` function in `_units.scss` to fix compiled CSS output


src/scss/tools/functions/_units.scss
```diff
@@ -68,6 +51,23 @@ $line-height-base : line-height(); // Lifted out of the variables file to fix ci
...
+
+        @if type-of($font-list) == 'list' {
+            @return line-height(nth($font-list, 1), nth($font-list, 2));
+        } @else {
+            @warn 'No line-height found as part of the $type map for #{ $font-size }';
+        }
+    }
+}
+
+$line-height-base : line-height(); // Lifted out of the variables file to fix circular reference issues
```
jamieomaguire, line 69: Sass doesn't seem to have hoisting so the line-height wasn't being compiled correctly. Hence why I've moved the order of these around.

package.json
```diff
@@ -34,7 +34,9 @@
   },
   "dependencies": {
     "@justeat/pie-design-tokens": "3.0.0",
-    "include-media": "eduardoboucas/include-media#2.0-release"
+    "glob": "8.0.3",
```
jamieomaguire, line 37: I was getting a lot of extraneous-dependencies which is why I moved these to regular dependencies. Happy to move back, will just need to add some Eslint overrides to the erroring files

kevinrodrigues: What were the errors out of interest?

jamieomaguire: it basically said this should be a regular dependency whenever I required the module

jamieomaguire: to be clear it was an IDE error rather than a compilation one

src/scss/components/optional/_order-card.scss
```diff
@@ -119,7 +119,7 @@
 
     .c-orderCard-order {
         @include type.font-size(body-l, false);
-        line-height: line-height(16px, 24px);
+        line-height: functions.line-height(16px, 24px);
```
kevinrodrigues, line 122: Thanks for fixing these ⭐

src/test/scss/data/fozzie-snapshot.scss
```diff
@@ -0,0 +1,128 @@
+@use 'fozzie' as f;
+
+// This file should call every part of Fozzie.
```
kevinrodrigues, line 3: Do we need to add something to the readme about unit testing? I assume every time a new component or style is added we need to modify this file? Does the order of the includes here break the snapshots too?

jamieomaguire: Happy to update the readme. The order would absolutely break the snapshot as we're changing the outputted CSS :)

package.json
kevinrodrigues: it maybe worth reverting the change then?

jamieomaguire: I'm happy to, just unsure if it's a legit concern as it seems to go against our eslint config

src/test/scss/data/fozzie-snapshot.scss
jamieomaguire: The idea was to get started with a big snapshot of everything. Unit tests could get a bit more granular but I'd be happy to do individual snapshots on parts of the library in theory too. I think for this ticket, 1 snapshot of everything is appropriate though

jamieomaguire: happy to create additional tickets

package.json
DamianMullins: If it's not required by projects consuming this package to use it then I'd say it belongs in `devDependencies`.

DamianMullins: Also, moving `sass` into this section will force consumers to pull that version locally.

src/test/scss/data/fozzie-snapshot.scss
kevinrodrigues: Sounds good 👍 I think granular would make it much easier to understand / add too in future.

package.json
jamieomaguire: will move it back now 👍

kevinrodrigues (approved):

README.md
```diff
@@ -67,3 +67,17 @@ Once you have imported fozzie into your Sass, you'll have access to the fozzie v
...
+2. Snapshot testing
+
+
+### Unit Testing
+
+We use a library called `sass-true` to enable writing unit tests for sass functions and mixins. These tests live in `src/test/scss/unit-tests`. We should use these as a means of documentation for our functions and mixins, as well as ensuring that API regressions aren't introduced.
+
+### Snapshot Testing
+
+We use `Jest` to write snapshot tests of the compiled CSS for parts of Fozzie. These live in `src/test/scss/snapshot-tests`. Snapshot tests provide a means of ensuring that no unexpected styles will be introduced for consumers of the library. They can also be used to ensure that the compiled CSS is 100% valid, or to make sure there's simply nothing unexpected being rendered. To write these tests, we can either import an SCSS file, or write a line of SCSS we'd like to test and compile it using the `compileToCss.js` module. We can then use Jest to snapshot test the outputted string.
```
ashleynolan, line 83:
```suggestion
We use `Jest` to write snapshot tests of the compiled CSS for parts of Fozzie. These live in `src/test/scss/snapshot-tests`. Snapshot tests provide a means of ensuring that no unexpected styles will be introduced for consumers of the library. They can also be used to ensure that the compiled CSS is 100% valid, or to make sure there's simply nothing unexpected being rendered.

To write these tests, we can either import an SCSS file or write a line of SCSS we'd like to test and compile it using the `compileToCss.js` module. We can then use Jest to snapshot test the outputted string.
```

src/scss/tools/functions/_units.scss
ashleynolan, line 69: Out of interest, is there a reason this needs to move down below the functions? Feels a bit hidden away

ashleynolan: Ah ok – so it's a hoisting issue essentially. No worries was just wondering!

kevinrodrigues (approved):

